PR #1, opened 2021-05-05
WIP: A first stab at analytic opacities

Yurlungur:
This is a first pass at analytic opacities. I'm not done, and I also haven't fully tested the machinery. But I wanted to get thoughts on what I have so far. In particular, I wanted your opinion on the function signatures, the structure of the variant, and the emphasis on header-only calls.

Note the function signatures don't expose Ye but they do expose radiation type. My reasoning for allowing this is that in principle one could run, e.g., neutrinos and photons at the same time, or some other kind of radiation. I'm not necessarily attached to this. The radiation type could be moved into the lambda and hidden, though it would be a little gross.

### DIFF
--- a/opacities/gray_opacity.hpp
+++ b/opacities/gray_opacity.hpp
@@ -1,0 +1,90 @@
+// ======================================================================
+// © 2021. Triad National Security, LLC. All rights reserved.  This
+// program was produced under U.S. Government contract
+// 89233218CNA000001 for Los Alamos National Laboratory (LANL), which
+// is operated by Triad National Security, LLC for the U.S.
+// Department of Energy/National Nuclear Security Administration. All
+// rights in the program are reserved by Triad National Security, LLC,
+// and the U.S. Department of Energy/National Nuclear Security
+// Administration. The Government is granted for itself and others
+// acting on its behalf a nonexclusive, paid-up, irrevocable worldwide
+// license in this material to reproduce, prepare derivative works,
+// distribute copies to the public, perform publicly and display
+// publicly, and to permit others to do so.
+// ======================================================================
+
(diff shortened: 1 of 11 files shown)


Yurlungur:
Pinging @jdolence @dholladay00 @brryan

opacities/gray_opacity.hpp
```diff
@@ -0,0 +1,90 @@
...
+#include <opac-utils/physical_constants.hpp>
+#include <ports-of-call/portability.hpp>
+
+#include "thermal_distributions.hpp"
+
+namespace singularity {
+
+template <typename ThermalDistribution, int NSPECIES> class GrayOpacity {
+public:
+  GrayOpacity(const Real kappa) : kappa_(kappa) {}
```
brryan, line 33: kappa is often an opacity (dimensions cm^2 g^-1) whereas here kappa is an absorption coefficient (dimensions cm^-1). My fault for having done this in earlier code I think. If we're passing in opacities should not `EmissivityPerNuOmega` et al. then return `rho*kappa_*Bnu`?

You could add a separate class, `GrayAbsortionCoefficient`, for the most extreme approximators who want constant optical depth over some distance regardless of material density

brryan: On second thought implementing `GrayAbsorptionCoefficient` would probably be a waste of time.

Yurlungur: >  If we're passing in opacities should not EmissivityPerNuOmega et al. then return `rho*kappa_*Bnu`?

Good catch. This is a bug. Thanks. I don't think we need a `GrayAbsorptionCoefficient` though.

Yurlungur:
Superseded by #3.

PR #2, opened 2021-05-10
Add physical constants

brryan:
In lieu of a potential `singularity-constants` package, I'm adding the constants file I was going to use here.

### DIFF
--- a/constants/constants.hpp
+++ b/constants/constants.hpp
@@ -1,0 +1,235 @@
+// ======================================================================
+// © 2021. Triad National Security, LLC. All rights reserved.  This
+// program was produced under U.S. Government contract
+// 89233218CNA000001 for Los Alamos National Laboratory (LANL), which
+// is operated by Triad National Security, LLC for the U.S.
+// Department of Energy/National Nuclear Security Administration. All
+// rights in the program are reserved by Triad National Security, LLC,
+// and the U.S. Department of Energy/National Nuclear Security
+// Administration. The Government is granted for itself and others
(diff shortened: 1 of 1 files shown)


constants/constants.hpp
```diff
@@ -0,0 +1,235 @@
...
+// and the U.S. Department of Energy/National Nuclear Security
+// Administration. The Government is granted for itself and others
+// acting on its behalf a nonexclusive, paid-up, irrevocable worldwide
+// license in this material to reproduce, prepare derivative works,
+// distribute copies to the public, perform publicly and display
+// publicly, and to permit others to do so.
+// ======================================================================
+
+#ifndef CONSTANTS_CONSTANTS_
+#define CONSTANTS_CONSTANTS_
```
Yurlungur: maybe this include guard should be something like `SINGULARITY_CONSTANTS_CONSTANTS_` or something just in case this guard signature is used elsewhere, which I can imagine happening.

Yurlungur (approved):
Looks good. One minor nitpick and then lets get this in.

constants/constants.hpp
brryan: Yeah good point, changed. I also `#undef`'d `CONSTANTS_ERROR` and `UNDEFINED_ERROR` at the end of the file. Should the exceptions-disabled version of `CONSTANTS_ERROR` call `exit()` as well as `printf()`?

Yurlungur: nvcc yells at me for `exit`. So I don't think we should add it.

Yurlungur:
Merging

PR #3, opened 2021-05-18
WIP: A second pass at analytic opacities

Yurlungur:
Based on feedback from my last attempt, PR #1, I attempt a different version of opacities, where the photon and neutrino versions are split. I manage to share some code for the variant object with some variatic templates.

@jdolence @brryan what do you think? If you like this, let's close PR #1 and merge this in. Then I will start thinking about tabulated opacities.


brryan (approved):
looks good!

singularity-opac/neutrinos/thermal_distributions_neutrinos.hpp
```diff
@@ -0,0 +1,68 @@
...
+  }
+  PORTABLE_INLINE_FUNCTION
+  Real ThermalNumberDistribution(const Real temp, const Real Ye, const RadiationType type,
+                                 Real *lambda = nullptr) const {
+    constexpr Real zeta3 = 1.20206;
+    return 12. * pow(pc::kb, 3) * M_PI * NSPECIES * pow(temp, 3) * zeta3 /
+           (pow(pc::c, 2) * pow(pc::h, 3));
+  }
+  template <typename Emissivity>
+  PORTABLE_INLINE_FUNCTION Real OpacityFromKirkhoff(const Emissivity &J, const Real rho,
```
brryan: Technically `AbsorptionCoefficientFromKirkhoff`, not `Opacity`

Yurlungur: Good call. I'll change that.

singularity-opac/photons/gray_opacity_photons.hpp
```diff
@@ -0,0 +1,81 @@
...
+    return rho * kappa_ * B;
+  }
+
+  PORTABLE_INLINE_FUNCTION
+  Real NumberEmissivity(const Real rho, const Real temp, Real *lambda = nullptr) const {
+    return kappa_ * dist_.ThermalNumberDistribution(temp, lambda);
+  }
+
+ private:
+  Real kappa_; // absorption coefficient. Units of 1/cm
```
brryan: Should we just call this an opacity with units cm^2/g? Looks like `rho` is always available to make that conversion.

singularity-opac/neutrinos/thermal_distributions_neutrinos.hpp
Yurlungur: Fixed.

singularity-opac/photons/gray_opacity_photons.hpp
Yurlungur: Yes. This should be an opacity. I'm multiplying everything by rho everywhere. I'll change the comment.

dholladay00:
For photons, are there ways to handle Planck vs Rosseland weightings? Is the `pernu` naming in Gray just for consistency? Sometimes (even in multigroup cases), one needs frequency integrated quantities, that path might be another one for gray.

Yurlungur:
> For photons, are there ways to handle Planck vs Rosseland weightings? Is the `pernu` naming in Gray just for consistency? Sometimes (even in multigroup cases), one needs frequency integrated quantities, that path might be another one for gray.

I haven't thought very carefully about that. The `PerNu` naming in Gray is just for consistency, yes. What I've put here isn't binned into groups at all. I think a per group opacity would be another function. We could template on weighting... unless both would be needed in the same simulation?

PR #6, opened 2021-05-20
Prototype API for doing per-cell opacities before looping over particles

Yurlungur:
This is a prototype API for doing per-cell opacities before looping over particles. This also fits in nicely with doing per-group opacities that might be, e.g., Planck- or Rosalind-weighted.

This is a work in progress, but I think feedback from interested parties would be good at this point. Pinging @jdolence @brryan @dholladay00 

## Changes

- I implement a templated function that fills an `Indexer` object (which must have an `operator[]`) with some opacity quantity on a grid of frequencies
- The `Indexer` can be anything... and it can have interpolation capabilities if this is useful
- I attempted to test this with linear-linear, log-linear, and log-chebyshev interpolation
- I had to split `opac_variant` into a separate class for the neutrino and photon opacities.

## Limitations

- Currently it's only implemented for some of the opacity functions. Needs to be implemented for all of them
- I've only tested it on gray opacities. I still need to produce non-gray to test this out on.


singularity-opac/chebyshev/chebyshev.hpp
```diff
@@ -0,0 +1,245 @@
...
+namespace singularity {
+namespace chebyshev {
+
+// Inverse Vandermonde matrices computed via Mathematica
+// Mathematica code:
+// GetChebNodes[n_] := -Table[Cos[(2*k - 1)/(2*n) \[Pi]], {k, 1, n}]
+// GetS2C[n_] := Transpose[Table[ChebyshevT[k, #] & /@ GetChebNodes[n],
+//                               {k, 0, n - 1}]]
+// GetS2C[21] // N // Inverse // Chop (* for 21 modes *)
+constexpr Real Vandermonde3[3][3] = {
```
jdolence: is there any reason to care about precision in these matrices?

singularity-opac/chebyshev/chebyshev.hpp
```diff
@@ -0,0 +1,245 @@
...
+    points[k - 1] = FromUnitCell(tk, xmin, xmax);
+  }
+}
+
+// TODO(JMM): Should this be templated? Or Recursive? What's faster?
+// Templates, iteration, or recursion?
+// Recursion relation for Chebyshev polynomials of the first kind:
+// T_0(x) = 1
+// T_1(x) = x
+// T_i(x) = 2*x*T_{i-1}(x) - T_{i-2}(x)
```
jdolence, line 69: tail recursion should be pretty efficient, but I have no idea if that holds on device.

singularity-opac/photons/gray_opacity_photons.hpp
```diff
@@ -49,19 +49,48 @@ class GrayOpacity {
                                                    lambda);
   }
 
+  template <typename FrequencyIndexer, typename DataIndexer>
+  PORTABLE_INLINE_FUNCTION void AbsorptionCoefficientPerNu(
```
jdolence, line 53: why does this gray opacity make repeated calls to fill in opacities for different frequency bins?

jdolence (approved):
Not the most thorough review, but it looks pretty good to me.  Still have a few questions, but nothing that needs to be resolved right now.

One question was whether there should be some super object that holds opacities + an indexer?

Yurlungur:
> Not the most thorough review, but it looks pretty good to me. Still have a few questions, but nothing that needs to be resolved right now.

Thanks, @jdolence!

> One question was whether there should be some super object that holds opacities + an indexer?

I thought about that and I decided against. The main reason I designed the indexers the way I did is that I'm imagining that these might be implemented on the host code side. For example, if the per-cell quantities are stored somehow on the host code, I wanted the host code to have the freedom to index into them as it liked. I could be wrong though. This is definitely worth thinking about more.

singularity-opac/chebyshev/chebyshev.hpp
Yurlungur: Yes. Your answer will only be as good as the precision here. I should probably be using `N[14]` and `Chop[14]` in the Mathematica code to make sure these have the required precision.

singularity-opac/chebyshev/chebyshev.hpp
Yurlungur: I think I'm going to leave it as iteration, mainly because two-element recursion has complexity O(2^n), whereas iteration has complexity O(n).

Yurlungur: But this is worth experimenting with to see if it matters. The recursion function is much simpler.

singularity-opac/photons/gray_opacity_photons.hpp
Yurlungur: As proof of principle. It should just return rho*kappa in this case. I will optimize this later when I have non-gray opacities to work with.

singularity-opac/chebyshev/chebyshev.hpp
Yurlungur: Fixed.

PR #10, opened 2021-06-29
Neutrino Tables

Yurlungur:
A first pass at neutrino tables. For testing purposes, I just fill the tables with gray opacities. Once I'm certain that works and happy with that machinery, I will write some code to generate some real tables in Spiner format.

That said, I think this is ready for feedback from interested parties. @jdolence @brryan I would like your feedback in particular. However, @dholladay00 and @mauneyc-LANL your feedback would be welcome too.

Many changes here actually happened in Spiner. In particular, for performance reasons, I needed to add [two](https://github.com/lanl/spiner/pull/8) [new](https://github.com/lanl/spiner/pull/10) interpolation operators, which mix interpolation and table indexing. This is because for neutrinos, we have five dimensions:
- density
- temperature
- electron fraction
- species
- frequency/energy

where species is a discrete index and the others are continuous. There are many ways of organizing the data here. I chose the ordering above, where density is slowest moving and frequency is fastest. This reflects the most common loop pattern, I expect... or at least one of two common loop patterns, the other with species fastest moving and frequency less fast. This may offer optimizations in the future where, for example, we loop over cells, interpolate on to a new, per-cell energy grid, and then use that new grid for per-particle interpolation.

Anyway, please let me know what you think.


Yurlungur:
Now relies on #11 so we shouldn't merge until that's in.

Yurlungur:
Interesting discovery: the log10 tricks we played before may not work with emissivities/opacities: the mantissa is too large.

Yurlungur:
On CPU with a pre-filled gray opacity, I am now confident the reader is working.

Yurlungur:
I discovered a strange interaction with passing by reference and the lambdas used in `mpark::visit`. Code now captures by reference for those lambdas to avoid mangling types.

Yurlungur:
@brryan @jdolence @mauneyc-LANL @dholladay00 I believe this is ready for review.

cmake/SetupDeps.cmake
```diff
@@ -3,19 +3,29 @@ include(FeatureSummary)
 # Setup CUDAToolkit
 # - provideds CUDA::toolkit
 #=======================================
-find_package(CUDAToolkit QUIET)
+if (NOT TARGET CUDA::toolkit)
```
brryan: Not sure I understand this diff... `main` seems to already have this logic that github claims will be added by this PR

brryan (approved):
LGTM, apart from the one github question

cmake/SetupDeps.cmake
Yurlungur: Oh that's weird. I have no idea what's going on here. Let me try and manually merge with main.

Yurlungur: ok I fixed it. Problem was I needed to merge main into the branch.

Yurlungur:
@jdolence do you want to have a look too or should I just merge now that @brryan has given the ok?

Yurlungur:
@jdolence had his chance. Merging when tests pass.

PR #11, opened 2021-07-01
Add include guards and fix cmake_dependent_option

Yurlungur:
Several cmake tweaks:
- Include guards so that when `singularity-opac` is built downstream, it only calls `FindPackage` if the needed targets are not already provided by a parent package. This is needed to keep package consistency for in-tree builds, so all subpackages use the same libraries. Thanks to @brryan for finding this issue.
- Added a bit more verbose messaging for these target checks
- Fixed an inverted argument list in `cmake_dependent_option` to prevent `cmake` from trying to link MPI when MPI is not found.

### DIFF
--- a/cmake/SetupDeps.cmake
+++ b/cmake/SetupDeps.cmake
@@ -3,19 +3,29 @@ include(FeatureSummary)
 # Setup CUDAToolkit
 # - provideds CUDA::toolkit
 #=======================================
-find_package(CUDAToolkit QUIET)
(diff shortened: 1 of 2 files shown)


brryan (approved):
LGTM! And this works for my environment without me having to `spack unload kokkos`

Yurlungur:
@mauneyc-LANL any objections before I pull the trigger on this?

Yurlungur:
Pulling the trigger.

PR #12, opened 2021-07-15
Add singularity-opac::flags cmake target

brryan:
To avoid relative include paths for headers in dependencies like ports-of-call, we want to add a `singularity-opac::flags` cmake target like how `singularity-eos` already works. This PR adds that feature. Not all the details in the `singularity-eos` implementation are included here (e.g. 
```cmake
    if(SINGULARITY_HIDE_MORE_WARNINGS)                                                               
      target_compile_options(                                                                        
        singularity-eos::flags                                                                       
         INTERFACE # Generator expression shamelessly copied from EAP                                
         "$<$<COMPILE_LANGUAGE:CXX>:--expt-relaxed-constexpr;>"                                      
           )                                                                                         
    else()                                                                                           
      target_compile_options(                                                                        
        singularity-eos::flags                                                                       
         INTERFACE # Generator expression shamelessly copied from EAP                                
         "$<$<COMPILE_LANGUAGE:CXX>:--expt-relaxed-constexpr;-Xcudafe;--diag_suppress=esa_on_defaulted_function_ignored;>"
           )                                                                                         
    endif()
```
because I'm quickly getting out of my depth, but I think this is a good starting point.

### DIFF
--- a/singularity-opac/neutrinos/opac_neutrinos.hpp
+++ b/singularity-opac/neutrinos/opac_neutrinos.hpp
@@ -16,7 +16,7 @@
 #ifndef SINGULARITY_OPAC_NEUTRINOS_OPAC_NEUTRINOS_
 #define SINGULARITY_OPAC_NEUTRINOS_OPAC_NEUTRINOS_
 
-#include <variant/include/mpark/variant.hpp>
+#include "../../utils/variant/include/mpark/variant.hpp"
 
 #include <singularity-opac/neutrinos/gray_opacity_neutrinos.hpp>
 #include <singularity-opac/neutrinos/neutrino_variant.hpp>
--- a/singularity-opac/neutrinos/opac_neutrinos.hpp
+++ b/singularity-opac/neutrinos/opac_neutrinos.hpp
@@ -16,7 +16,7 @@
 #ifndef SINGULARITY_OPAC_NEUTRINOS_OPAC_NEUTRINOS_
 #define SINGULARITY_OPAC_NEUTRINOS_OPAC_NEUTRINOS_
 
-#include "../../utils/variant/include/mpark/variant.hpp"
+#include <variant/include/mpark/variant.hpp>
 
 #include <singularity-opac/neutrinos/gray_opacity_neutrinos.hpp>
 #include <singularity-opac/neutrinos/neutrino_variant.hpp>
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,6 +48,11 @@ add_library(${PROJECT_NAME} INTERFACE IMPORTED GLOBAL)
 # so we can use with add_subdirectory
 add_library(${PROJECT_NAME}::${PROJECT_NAME} ALIAS ${PROJECT_NAME})
 
+add_library (singularity-opac::flags INTERFACE IMPORTED GLOBAL)
+target_include_directories(singularity-opac::flags
+                           INTERFACE
+                           $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>)
+
 include (SetupDeps)
 include (SetupOptions)
 include (SetupCompilers)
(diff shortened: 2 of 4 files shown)


singularity-opac/neutrinos/opac_neutrinos.hpp
```diff
@@ -16,7 +16,7 @@
 #ifndef SINGULARITY_OPAC_NEUTRINOS_OPAC_NEUTRINOS_
 #define SINGULARITY_OPAC_NEUTRINOS_OPAC_NEUTRINOS_
 
-#include <variant/include/mpark/variant.hpp>
+#include "../../utils/variant/include/mpark/variant.hpp"
```
Yurlungur: Should this relative path be changed? I think the global path should still work?

Yurlungur (approved):
Thanks for this, @brryan ! I imagine @mauneyc-LANL may at some point want to move these settings into different parts of the cmake machinery (e.g., different files). However, I would say we get this in, so that we have a working capability, and then change it later as needed to clean things up.

Just one issue below, which I'd like to see changed before we merge.

singularity-opac/neutrinos/opac_neutrinos.hpp
Yurlungur: Ah you already fixed it. :+1:

brryan: already fixed :)

mauneyc-LANL:
This may be unnecessary; in my updates to cmake, I changed the flags interface to:

```
add_library(${PROJECT_NAME} INTERFACE IMPORTED GLOBAL)
add_library(${PROJECT_NAME}::${PROJECT_NAME} ALIAS ${PROJECT_NAME})
```

so that `singularity-opac::singularity-opac` would be the interface to use. The XL flags would be brought in, etc. This is mostly boiler-plate, and nothing other than taste really forces this to be done this way.

another possibility would be to add

```
add_library(${PROJECT_NAME}::flags ALIAS ${PROJECT_NAME})
```

to maintain the old interface as `singularity-opac::flags`

Yurlungur:
Ah... I see. @brryan can you try using `singularity-opac::singularity-opac`? If that works, we can remove the `::flags` interface.

brryan:
@Yurlungur @mauneyc-LANL I tried this (did I do this right?):

```cmake
target_link_libraries(phoebus PRIVATE parthenon                                                      
  singularity-eos::flags                                                                             
  singularity-eos                                                                                    
  singularity-opac::singularity-opac                                                                 
  #singularity-opac::flags                                                                           
  #singularity-opac                                                                                  
  $<$<BOOL:${PHOEBUS_ENABLE_HDF5}>:singularity-eos::hdf5> # TODO(BRR) what do I do here              
)  
```

but I get linking errors:

```
CMakeFiles/phoebus.dir/radiation/monte_carlo.cpp.o: In function `double singularity::neutrinos::FermiDiracDistributionNoMu<3>::AbsorptionCoefficientFromKirkhoff<singularity::neutrinos::TophatEmissivity<singularity::neutrinos::FermiDiracDistributionNoMu<3> > >(singularity::neutrinos::TophatEmissivity<singularity::neutrinos::FermiDiracDistributionNoMu<3> > const&, double, double, double, singularity::RadiationType, double, double*) const':
/home/brryan/github/phoebus/external/singularity-opac/singularity-opac/neutrinos/thermal_distributions_neutrinos.hpp:58: undefined reference to `singularity::neutrinos::FermiDiracDistributionNoMu<3>::EPS'
/home/brryan/github/phoebus/external/singularity-opac/singularity-opac/neutrinos/thermal_distributions_neutrinos.hpp:59: undefined reference to `singularity::neutrinos::FermiDiracDistributionNoMu<3>::EPS'
collect2: error: ld returned 1 exit status
src/CMakeFiles/phoebus.dir/build.make:565: recipe for target 'src/phoebus' failed
make[2]: *** [src/phoebus] Error 1
CMakeFiles/Makefile2:1784: recipe for target 'src/CMakeFiles/phoebus.dir/all' failed
make[1]: *** [src/CMakeFiles/phoebus.dir/all] Error 2
Makefile:181: recipe for target 'all' failed
make: *** [all] Error 2
```

brryan:
Hang on, sorry, this could be due to some other pollution in my current checkout

brryan:
OK forget what I said before, I *can* compile the code with only `singularity-opac::singularity-opac` :+1:

mauneyc-LANL:
Great! I'll put together a PR without `::flags` code

PR #13, opened 2021-07-15
Add inline to fix multiple definitions errors

brryan:
in `singularity-opac/base/radiation_types.hpp` there are two functions `int RadType2Idx(RadiationType type)` and `RadiationType Idx2RadType(int i)` which (I think generally, but definitely for my use case) give multiple definition errors when compiling/linking code that uses `singularity-opac` in multiple places. I here add `inline` qualifiers to these functions to fix that issue.


Yurlungur (approved):
Whoopsies. Good catch, thanks @brryan !

PR #14, opened 2021-07-20
Add mp

brryan:
`mp` missing from tophat number emissivity formula for neutrinos (as in nubhlight paper)


brryan:
This is not a bug, I was just confused. Talking with @Yurlungur, let's leave the mp out so this is indeed a number emissivity and not a mass emissivity.

PR #16, opened 2021-07-23
Add angle-averaged absorption coefficients

brryan:
Add `AngleAveragedAbsorptionCoefficientPerNu`. Same units as absorption coefficients (1/length) but ignores potential theta, phi structure in the `EmissivityPerNuOmega`.

### DIFF
--- a/singularity-opac/neutrinos/spiner_opac_neutrinos.hpp
+++ b/singularity-opac/neutrinos/spiner_opac_neutrinos.hpp
@@ -86,7 +86,7 @@ class SpinerOpacity {
     lalphanu_.setRange(2, YeMin, YeMax, NYe);
     lalphanu_.setRange(3, lTMin, lTMax, NT);
     lalphanu_.setRange(4, lRhoMin, lRhoMax, NRho);
-    lavgdalphanu_.copyMetadata(lalphanu_);
+    lAlphanu_.copyMetadata(lalphanu_);
     ljnu_.copyMetadata(lalphanu_);
 
     // set metadata for lJ and lJYe
@@ -117,9 +117,9 @@ class SpinerOpacity {
               Real E = fromLog_(lE);
               Real nu = MeV2Hz * E;
               Real alpha = std::max(
-                  opac.AbsorptionCoefficientPerNu(rho, T, Ye, type, nu), 0.0);
+                  opac.AbsorptionCoefficient(rho, T, Ye, type, nu), 0.0);
               lalphanu_(iRho, iT, iYe, idx, ie) = toLog_(alpha);
-              lavgdalphanu_(iRho, iT, iYe, idx, ie) = lalphanu_(iRho, iT, iYe, idx, ie);
+              lAlphanu_(iRho, iT, iYe, idx, ie) = lalphanu_(iRho, iT, iYe, idx, ie);
               Real j = std::max(opac.EmissivityPerNuOmega(rho, T, Ye, type, nu),
                                 0.0);
               ljnu_(iRho, iT, iYe, idx, ie) = toLog_(j);
@@ -132,10 +132,11 @@ class SpinerOpacity {
 
   // DataBox constructor. Note that this constructor *shallow* copies
   // the databoxes, so they must be managed externally.
-  SpinerOpacity(const Spiner::DataBox &lalphanu, const Spiner::DataBox ljnu,
-                const Spiner::DataBox lJ, const Spiner::DataBox lJYe)
+  SpinerOpacity(const Spiner::DataBox &lalphanu, const Spiner::DataBox &lAlphanu,
+                const Spiner::DataBox ljnu, const Spiner::DataBox lJ,
+                const Spiner::DataBox lJYe)
       : memoryStatus_(impl::DataStatus::OnHost), lalphanu_(lalphanu),
-        lavgdalphanu_(lavgdalphanu), ljnu_(ljnu), lJ_(lJ), lJYe_(lJYe) {}
+        lAlphanu_(lAlphanu), ljnu_(ljnu), lJ_(lJ), lJYe_(lJYe) {}
 
 #ifdef SPINER_USE_HDF
   SpinerOpacity(const std::string &filename)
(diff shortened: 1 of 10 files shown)


Yurlungur:
@brryan this seems reasonable to me but I have a few questions. Will you then make `AbsorptionCoefficientPerNu` depend on angle? Should it be renamed `AbsorptionCoefficientPerNuOmega`?

Yurlungur (commented):
Also looks like you're missing changes to the tabulated one.

brryan:
> @brryan this seems reasonable to me but I have a few questions. Will you then make `AbsorptionCoefficientPerNu` depend on angle? Should it be renamed `AbsorptionCoefficientPerNuOmega`?

@Yurlungur yes I think in general `AbsorptionCoefficientPerNu` can now depend on angle, but I used this name because there isn't a 1/dOmega in the definition of this quantity as in the emissivity. And just like `EmissivityPerNuOmega`, angular parameters like `theta` would just be passed in through the `*lambda`, right?

If that doesn't work, or there are clearer possible names, I'm happy to change this though

Yurlungur:
> @Yurlungur yes I think in general `AbsorptionCoefficientPerNu` can now depend on angle, but I used this name because there isn't a 1/dOmega in the definition of this quantity as in the emissivity.

RIght. I guess my question is: Should there be a `1/dOmega` in the definition?

> And just like `EmissivityPerNuOmega`, angular parameters like `theta` would just be passed in through the `*lambda`, right?

Yeah... I guess that's currently how it's set up. Should that change? Should we add theta and phi to the definition of emissivity so that those parameters don't have to go through the `lambda`? What do you think?

brryan:
> RIght. I guess my question is: Should there be a `1/dOmega` in the definition?

I think no, since in both cases this is giving us the quantity j_nu / B_nu = <1/dx>, the only difference being either jnu or Jnu/4pi for the "j_nu" in the previous expression. Actually, because there's no 1/dnu in the definition of alpha either, should we even do `AbsorptionCoefficientPerNu` -> `AbsorptionCoefficient`?

> Yeah... I guess that's currently how it's set up. Should that change? Should we add theta and phi to the definition of emissivity so that those parameters don't have to go through the `lambda`? What do you think?

Personally, I'm happy with letting theta/phi be lambda parameters, since they won't always be needed (I know of no case where phi would be needed, actually).

Yurlungur:
> I think no, since in both cases this is giving us the quantity j_nu / B_nu = <1/dx>, the only difference being either jnu or Jnu/4pi for the "j_nu" in the previous expression

Works for me.

> Actually, because there's no 1/dnu in the definition of alpha either, should we even do AbsorptionCoefficientPerNu -> AbsorptionCoefficient?

Hmm good point. I would be okay with that.

> Personally, I'm happy with letting theta/phi be lambda parameters, since they won't always be needed (I know of no case where phi would be needed, actually).

Is that still true for photons? Should we change the photon API to include theta?

brryan:
> > Personally, I'm happy with letting theta/phi be lambda parameters, since they won't always be needed (I know of no case where phi would be needed, actually).
> 
> Is that still true for photons? Should we change the photon API to include theta?

Even for photons it depends on the physics. The emissivity for e-p bremsstrahlung on an isotropic distribution of electrons will have no theta dependence, for example

Yurlungur:
> Even for photons it depends on the physics. The emissivity for e-p bremsstrahlung on an isotropic distribution of electrons will have no theta dependence, for example

Yeah ok. How about we change `AbsorptionCoefficientPerNu` to `AbsorptionCoefficient` and then call it a day and merge this in?

brryan:
> Yeah ok. How about we change `AbsorptionCoefficientPerNu` to `AbsorptionCoefficient` and then call it a day and merge this in?

Sounds good. I'll fix the tabulated calls too, thanks for pointing those out.

Yurlungur:
@brryan ping me when it's ready for 2nd review

brryan:
@Yurlungur OK this is ready for re-review. I'm not sure it's showing diffs for `spiner` since it looks like `main` and this branch have the same spiner hash, `cadaf1e` (the latest).

singularity-opac/neutrinos/spiner_opac_neutrinos.hpp
```diff
@@ -130,17 +132,19 @@ class SpinerOpacity {
...
   // DataBox constructor. Note that this constructor *shallow* copies
   // the databoxes, so they must be managed externally.
-  SpinerOpacity(const Spiner::DataBox &lalphanu, const Spiner::DataBox ljnu,
-                const Spiner::DataBox lJ, const Spiner::DataBox lJYe)
+  SpinerOpacity(const Spiner::DataBox &lalphanu, const Spiner::DataBox &lAlphanu,
+                const Spiner::DataBox ljnu, const Spiner::DataBox lJ,
+                const Spiner::DataBox lJYe)
       : memoryStatus_(impl::DataStatus::OnHost), lalphanu_(lalphanu),
-        ljnu_(ljnu), lJ_(lJ), lJYe_(lJYe) {}
+        lAlphanu_(lAlphanu), ljnu_(ljnu), lJ_(lJ), lJYe_(lJYe) {}
```
Yurlungur: I'm having second thoughts about actually introducing a separate table, rather than just changing the API but keep only 4 tables. The Burrows tables don't look any different between Alpha and alpha. Maybe we should save space and only use one table. What do you think @brryan ?

Yurlungur (commented):
Sorry somehow I spaced on this. One (rather important) comment below. Otherwise LGTM.

singularity-opac/neutrinos/spiner_opac_neutrinos.hpp
brryan: You're right, I shouldn't add data just for prettiness reasons. Doubt we'll ever care about the distinction here. What do you think of this now?

Also wow I forgot about this PR for a long time.

Yurlungur (approved):
LGTM I'll merge in

singularity-opac/neutrinos/spiner_opac_neutrinos.hpp
Yurlungur: Cool thanks @brryan and no problem. :)

PR #17, opened 2021-07-27
Switch from constexpr to preproc for EPS for C++14 compatibility

brryan:
Downstream I'm compiling with C++14 which doesn't have inline variables. I was getting linking errors with the `constexpr` implementation of `EPS` as seen in the changes. This changeset works with my code and I claim it is relatively safe because `EPS` is `#undef`'d. If there's another way to get the `constexpr` approach to work in a header-only framework I'd be happy to do that instead though.


Yurlungur (approved):
Gross but acceptable. Merging.

PR #18, opened 2021-09-23
units modifier for opacities

Yurlungur:
Following up on the units modifier PR for `singularity-eos` here's one in `singularity-opac`.

### DIFF
--- a/singularity-opac/base/indexers.hpp
+++ b/singularity-opac/base/indexers.hpp
@@ -34,7 +34,7 @@ namespace indexers {
 template <typename T>
 struct Identity {
   Identity() = default;
-  Identity(T &t) : data_(std::forward<T>(t)) {}
+  Identity(T &&t) : data_(std::forward<T>(t)) {}
 
   PORTABLE_INLINE_FUNCTION
   Real &operator[](const int i) { return data_[i]; }
--- a/singularity-opac/base/indexers.hpp
+++ b/singularity-opac/base/indexers.hpp
@@ -31,6 +31,19 @@
 namespace singularity {
 namespace indexers {
 
+template <typename T>
+struct Identity {
+  Identity() = default;
+  Identity(T &t) : data_(std::forward<T>(t)) {}
+
+  PORTABLE_INLINE_FUNCTION
+  Real &operator[](const int i) { return data_[i]; }
+  PORTABLE_INLINE_FUNCTION
+  Real &operator[](const int i) const { return data_[i]; }
+
+  T data_;
+};
+
 class Linear {
  public:
   Linear() = default;
--- a/singularity-opac/neutrinos/non_cgs_neutrinos.hpp
+++ b/singularity-opac/neutrinos/non_cgs_neutrinos.hpp
@@ -1,0 +1,172 @@
+// ======================================================================
+// © 2021. Triad National Security, LLC. All rights reserved.  This
+// program was produced under U.S. Government contract
+// 89233218CNA000001 for Los Alamos National Laboratory (LANL), which
+// is operated by Triad National Security, LLC for the U.S.
+// Department of Energy/National Nuclear Security Administration. All
+// rights in the program are reserved by Triad National Security, LLC,
+// and the U.S. Department of Energy/National Nuclear Security
+// Administration. The Government is granted for itself and others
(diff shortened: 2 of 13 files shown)


Yurlungur:
This isn't done. I've only implemented it in neutrinos so far and not tested it yet. All that's forthcoming.

singularity-opac/base/indexers.hpp
```diff
@@ -31,6 +31,19 @@
 namespace singularity {
 namespace indexers {
 
+template <typename T>
+struct Identity {
+  Identity() = default;
+  Identity(T &t) : data_(std::forward<T>(t)) {}
```
mauneyc-LANL: Is this a copy or move constructor? You are passing an l-value reference, but forwarding it (which is a no-op on l-values). You can use `Identity(T&& t) : data_(std::forward<T>(t)) {}` to let `forward` choose copy or move, but this isn't idiomatic.

Yurlungur: Oops good catch.

singularity-opac/neutrinos/non_cgs_neutrinos.hpp
```diff
@@ -0,0 +1,189 @@
...
+#include <singularity-opac/base/opac_error.hpp>
+
+namespace singularity {
+namespace neutrinos {
+
+template <typename Opac>
+class NonCGSUnits {
+ public:
+  NonCGSUnits() = default;
+  NonCGSUnits(Opac &&opac, const Real time_unit, const Real mass_unit,
```
mauneyc-LANL, line 33: Similar to above, if using an 'universal reference' `&&` it should be `std::forward` ed ... that is, if the function accepts l-value and r-value (e.g. `NonCGSUnits( Opac{}, ...` will be moved). If it's explicitly meant to be moved, pass-by-value and initialize with `std::move`, e.g. `opac_(std::move(opac)`.

Yurlungur: Also good catch.

Yurlungur:
Now implemented for photons and tests pass. Thanks @mauneyc-LANL for the catches. Pinging @brryan for feedback.

singularity-opac/base/indexers.hpp
```diff
@@ -31,6 +31,19 @@
...
+  Identity(T &&t) : data_(std::forward<T>(t)) {}
+
+  PORTABLE_INLINE_FUNCTION
+  Real &operator[](const int i) { return data_[i]; }
+  PORTABLE_INLINE_FUNCTION
+  Real &operator[](const int i) const { return data_[i]; }
+
+  T data_;
+};
+
```
brryan, line 46: Is `Identity` used anywhere?

singularity-opac/neutrinos/non_cgs_neutrinos.hpp
```diff
@@ -0,0 +1,189 @@
...
+              const Real length_unit, const Real temp_unit)
+      : opac_(std::forward<Opac>(opac)), time_unit_(time_unit),
+        mass_unit_(mass_unit), length_unit_(length_unit), temp_unit_(temp_unit),
+        rho_unit_(mass_unit_ / (length_unit_ * length_unit_ * length_unit_)),
+        freq_unit_(1. / time_unit_),
+        inv_emiss_unit_(length_unit_ * time_unit_ * time_unit_ / mass_unit_),
+        inv_num_emiss_unit_(length_unit_ * length_unit_ * length_unit_ *
+                            time_unit_) {}
```
brryan, line 41: Possible someone would want to use a different "numbers" unit someday... like moles instead of actual number. Worth adding a `number_unit_`?

brryan (approved):
Just a few little comments, LGTM!

singularity-opac/base/indexers.hpp
Yurlungur: Uh no. It could be deleted. It's essentially a silly test.

singularity-opac/neutrinos/non_cgs_neutrinos.hpp
Yurlungur: Oh this is a good idea. Let's pass on this for now, but keep it in mind?

PR #19, opened 2022-01-27
Remove extraneous 4pi

brryan:
There was an extra `4*M_PI` in the denominator of the tophat `EmissivityPerNuOmega` -- note that the angle-integrated emissivities in the same file are *multiplied* by `4.*M_PI`


Yurlungur (approved):
Good catch thanks.

PR #20, opened 2022-03-08
Fix integrated distribution functions

brryan:
There was a missing 1/c factor in the integrated distribution functions for photons and neutrinos. The specific intensity `B_nu = dE / (dA dt dOmega dnu)`, so e.g. the energy density `u = dE / dV = 4 pi / c \int B_nu dnu`. This is consistent with `u = a_R T^4` for photons.

Also, I get a 16 instead of a 12 as the prefactor for the neutrino thermal number density, maybe I forgot some neutrino physics. Anyway here's my mathematica output:

image

image


Yurlungur (approved):
Good catches. I think these are both just mistakes on my part.

PR #24, opened 2022-03-15
Fix thermal distributions again

brryan:
I misunderstood what `ThermalDistributionOfT` was providing (and so I broke e.g. gray opacities in #20).

I also changed `ThermalNumberDistribution` to `ThermalNumberDistributionOfT` for symmetry.

Ultimately we want to provide both `I = \int I_{\nu} d\Omega d\nu` and `u = 1/c \int I_{\nu} d\Omega d\nu`. However, this is always just division by `c` so long as the particles are always moving at `c`. So let's not add the extra functions. I included a note in the `README` about this conversion.

I'm also adding to `README.md` to help avoid this kind of confusion going forward

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,15 +3,7 @@ singularity-opac
 
 [![Tests](https://github.com/lanl/singularity-opac/actions/workflows/tests.yml/badge.svg)](https://github.com/lanl/singularity-opac/actions/workflows/tests.yml)
 
-Performance Portable Opacity and Emissivity for simulation codes
-
-## Components
-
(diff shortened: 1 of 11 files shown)


Yurlungur:
I'm fine with letting users divide by C or adding another function. Dividing by C seems cleanest, though.

brryan:
OK @Yurlungur this is ready for review. Sorry for the thrashing.

Yurlungur (approved):
Thanks for added documentation, and sorry I didn't catch this initially.

PR #25, opened 2022-04-07
Add Planck and Rosseland means

### LINKED ISSUE #21: Add Planck and Rosseland mean opacities
Opacities are currently only available in per-frequency form.

brryan:
Frequency-integrated methods require opacities in some frequency-averaged sense. This PR provides Planck and Rosseland mean opacities through the interface. Not all existing opacity models are supported right now, I want to at least get `BRT` though.

### DIFF
--- a/singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
+++ b/singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
@@ -21,6 +21,7 @@
 #include <ports-of-call/portability.hpp>
 #include <singularity-opac/base/radiation_types.hpp>
 #include <singularity-opac/constants/constants.hpp>
+#include <spiner/databox.hpp>
 
 namespace singularity {
 namespace neutrinos {
@@ -35,9 +36,12 @@ using pc = PhysicalConstants<CGS>;
 class MeanOpacity {
  public:
   MeanOpacity() = default;
+  template <typename Opacity>
   MeanOpacity(const Opacity &opac, Real lRhoMin, Real lRhoMax, int NRho,
               Real lTMin, Real lTMax, int NT, Real YeMin, Real YeMax, int NYe,
               Real *lambda = nullptr) {
+    nlambda_ = opac.nlambda();
+
     lkappaPlanck_.resize(NRho, NT, NYe, NEUTRINO_NTYPES);
     // index 0 is the species and is not interpolatable
     lkappaPlanck_.setRange(1, YeMin, YeMax, NYe);
@@ -81,6 +85,37 @@ class MeanOpacity {
               kappaRosselandDenom +=
                   opac.DThermalDistributionOfTNuDT(T, type, nu) * nu * dlnu;
             }
+
+            // Trapezoidal rule
+            const Real nu0 = fromLog_(lnuMin);
+            const Real nu1 = fromLog_(lnuMax);
+            kappaPlanckNum -=
+                0.5 * 1. / rho *
+                (opac.AbsorptionCoefficient(rho, T, Ye, type, nu0, lambda) *
+                     opac.ThermalDistributionOfTNu(T, type, nu0) * nu0 +
+                 opac.AbsorptionCoefficient(rho, T, Ye, type, nu1, lambda) *
+                     opac.ThermalDistributionOfTNu(T, type, nu1) * nu1) *
+                dlnu;
+            kappaPlanckDenom -=
+                0.5 *
+                (opac.ThermalDistributionOfTNu(T, type, nu0) * nu0 +
+                 opac.ThermalDistributionOfTNu(T, type, nu1) * nu1) *
+                dlnu;
+            kappaRosselandNum -=
+                0.5 * rho *
+                (1. /
+                     opac.AbsorptionCoefficient(rho, T, Ye, type, nu0, lambda) *
+                     opac.DThermalDistributionOfTNuDT(T, type, nu0) * nu0 +
+                 1. /
+                     opac.AbsorptionCoefficient(rho, T, Ye, type, nu1, lambda) *
+                     opac.DThermalDistributionOfTNuDT(T, type, nu1) * nu1) *
+                dlnu;
+            kappaRosselandDenom -=
+                0.5 *
+                (opac.DThermalDistributionOfTNuDT(T, type, nu0) * nu0 +
+                 opac.DThermalDistributionOfTNuDT(T, type, nu1) * nu1) *
+                dlnu;
+
             Real lkappaPlanck = toLog_(kappaPlanckNum / kappaPlanckDenom);
             Real lkappaRosseland =
                 toLog_(1. / (kappaRosselandNum / kappaRosselandDenom));
@@ -92,8 +127,43 @@ class MeanOpacity {
     }
   }
 
+#ifdef SPINER_USE_HDF
+  SpinerOpacity(const std::string &filename)
+      : filename_(filename.c_str()), memoryStatus_(impl::DataStatus::OnHost) {
+    herr_t status = H5_SUCCESS;
+    hid_t file = H5Fopen(filename.c_str(), H5F_ACC_RDONLY, H5P_DEFAULT);
+    status += lkappaPlanck_.loadHDF(file, SP5::Opac::PlanckMeanOpacity);
+    status += lkappaRosseland_.loadHDF(file, SP5::Opac::RosselandMeanOpacity);
+    status += H5Fclose(file);
+
+    if (status != H5_SUCCESS) {
+      OPAC_ERROR("neutrinos::MeanOpacity: HDF5 error\n");
+    }
+  }
+
+  void Save(const std::string &filename) const {
+    herr_t status = H5_SUCCESS;
+    hid_t file =
+        H5Fcreate(filename.c_str(), H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);
+    status += lkappaPlanck_.saveHDF(file, SP5::Opac::PlanckMeanOpacity);
+    status += lkappaRosseland_.saveHDF(file, SP5::Opac::RosselandMeanOpacity);
+    status += H5Fclose(file);
+
+    if (status != H5_SUCCESS) {
+      OPAC_ERROR("neutrinos::MeanOpacity: HDF5 error\n");
+    }
+  }
+#endif
+
+  PORTABLE_INLINE_FUNCTION int nlambda() const { return nlambda_; }
+
+  PORTABLE_INLINE_FUNCTION void PrintParams() const {
+    printf("Mean opacity\n");
+  }
+
   MeanOpacity GetOnDevice() {
     MeanOpacity other;
+    other.nlambda_ = nlambda_;
     other.lkappaPlanck_ = Spiner::getOnDeviceDataBox(lkappaPlanck_);
     other.lkappaRosseland_ = Spiner::getOnDeviceDataBox(lkappaRosseland_);
     return other;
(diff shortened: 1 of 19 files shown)


Yurlungur:
It occurs to me that we could compute frequency-averaged opacities numerically if we want to. Should we? At runtime, we probably don't want to... it's probably slow... but we could imagine numerically computing them and then tabulating them in Opac Burrows or something.

brryan:
> It occurs to me that we could compute frequency-averaged opacities numerically if we want to. Should we? At runtime, we probably don't want to... it's probably slow... but we could imagine numerically computing them and then tabulating them in Opac Burrows or something.

Thinking more about this, I think in general we probably do want to do this -- even for "simple" analytic opacities like BRT I don't think the integrals are tractable. I still want to think a bit about where exactly this should be done... there's a "free parameter" in the number of temperature support points used, but maybe letting the downstream application set this isn't worth the cost of a potentially slow numerical integral at initialization. 

EDIT: I think these integrals have to be computed when the opacity object is created because they depend on the arguments passed to the constructor. We could create a separate `mean_opac` class (to avoid computing mean opacities if you don't need them) that takes an `opac` object and some resolution parameters and tabulates the mean opacities.

Yurlungur:
> EDIT: I think these integrals have to be computed when the opacity object is created because they depend on the arguments passed to the constructor. We could create a separate mean_opac class (to avoid computing mean opacities if you don't need them) that takes an opac object and some resolution parameters and tabulates the mean opacities.

I like this idea.

brryan:
@Yurlungur OK I think this is in an OK place now for review, and merge if you're happy with it. Works with non-CGS units, but without the explicit `Variant` class from the regular opacities. No photon support yet.

singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
```diff
@@ -0,0 +1,212 @@
...
+#include <singularity-opac/base/radiation_types.hpp>
+#include <singularity-opac/constants/constants.hpp>
+#include <spiner/databox.hpp>
+
+namespace singularity {
+namespace neutrinos {
+
+using pc = PhysicalConstants<CGS>;
+
+#define EPS (10.0 * std::numeric_limits<Real>::min())
```
brryan, line 32: The old `#define EPS (10.0 * std::numeric_limits<Real>::epsilon())` doesn't work for small numbers, e.g. a typical neutrino opacity of `1.e-20 cm^2/g`. I changed this everywhere.

Yurlungur: That seems fine. Is there a reason we did `#define` and not `constexpr Real`? The latter seems better...

Yurlungur: also should we get rid of the 10?

singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
```diff
@@ -0,0 +1,212 @@
...
+            lkappaPlanck_(iRho, iT, iYe, idx) = lkappaPlanck;
+            lkappaRosseland_(iRho, iT, iYe, idx) = lkappaRosseland;
+          }
+        }
+      }
+    }
+  }
+
+#ifdef SPINER_USE_HDF
+  SpinerOpacity(const std::string &filename)
```
Yurlungur: This should be `MeanOpacity`. Might be worth testing this capability works.

singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
```diff
@@ -0,0 +1,212 @@
...
+    hid_t file =
+        H5Fcreate(filename.c_str(), H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);
+    status += lkappaPlanck_.saveHDF(file, SP5::Opac::PlanckMeanOpacity);
+    status += lkappaRosseland_.saveHDF(file, SP5::Opac::RosselandMeanOpacity);
+    status += H5Fclose(file);
+
+    if (status != H5_SUCCESS) {
+      OPAC_ERROR("neutrinos::MeanOpacity: HDF5 error\n");
+    }
+  }
```
Yurlungur: Nice. Caching mean opacities in file seems useful.

singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
```diff
@@ -0,0 +1,212 @@
...
+#endif
+
+  PORTABLE_INLINE_FUNCTION int nlambda() const { return nlambda_; }
```
Yurlungur: None of the mean opacity functions have a `lambda` parameter. I would either add one, or remove this call. I guess it's worth thinking about whether we want a `lambda` input param to these functions. I suspect not, but I don't know.

singularity-opac/neutrinos/non_cgs_neutrinos.hpp
```diff
@@ -234,6 +242,58 @@ class NonCGSUnits {
...
+    // thus multiplication converts (1/cm) to unit system.
+    return alpha * length_unit_;
+  }
+
+ private:
+  MeanOpac mean_opac_;
+  Real time_unit_, mass_unit_, length_unit_, temp_unit_;
+  Real rho_unit_;
+};
+
```
Yurlungur, line 296: Is this class needed? Could you put the units directly in the mean opacity class? Do you think we'll be adding more mean opacity types? It doesn't matter much either way, I guess. But the modifier pattern might not be needed for mean opacities. I leave it up to you.

singularity-opac/neutrinos/spiner_opac_neutrinos.hpp
```diff
@@ -61,7 +61,7 @@ enum class DataStatus { Deallocated, OnDevice, OnHost };
 template <typename ThermalDistribution>
 class SpinerOpacity {
  public:
-  static constexpr Real EPS = 10.0 * std::numeric_limits<Real>::epsilon();
+  static constexpr Real EPS = 10.0 * std::numeric_limits<Real>::min();
```
Yurlungur, line 64: should we remove the 10?

singularity-opac/neutrinos/thermal_distributions_neutrinos.hpp
```diff
@@ -27,7 +27,7 @@ namespace neutrinos {
 
 using pc = PhysicalConstants<CGS>;
 
-#define EPS (10.0 * std::numeric_limits<Real>::epsilon())
+#define EPS (10.0 * std::numeric_limits<Real>::min())
```
Yurlungur, line 30: constexpr?

Yurlungur: Maybe `EPS` should be defined in `neutrinos/base` or something, so we don't have the same line floating around everywhere.

singularity-opac/photons/gray_opacity_photons.hpp
```diff
@@ -43,37 +43,37 @@ class GrayOpacity {
   inline void Finalize() noexcept {}
 
   PORTABLE_INLINE_FUNCTION
-  Real AbsorptionCoefficientPerNu(const Real rho, const Real temp,
-                                  const Real nu, Real *lambda = nullptr) const {
+  Real AbsorptionCoefficient(const Real rho, const Real temp, const Real nu,
+                             Real *lambda = nullptr) const {
```
Yurlungur, line 47: Why this change? Do we no longer include the bolumetric/gray absorption coefficient in each opacity class?

Yurlungur (changes requested):
Overall looks good. A few comments below.

singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
brryan: Well the `lambda` parameter can still be provided to the constructor that calls the non-mean opacities. It's just that lambda is assumed constant for all `rho`, `T`, `Ye`, `nu`.

singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
brryan: Nice catch! Cmake not currently working with HDF5 on for me unfortunately. Should I upgrade the CI to test with HDF5?

singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
brryan: :+1:

singularity-opac/neutrinos/non_cgs_neutrinos.hpp
brryan: I'd rather leave it this way so one can get CGS units back by default without having to think about a units modifier at all.

singularity-opac/neutrinos/spiner_opac_neutrinos.hpp
brryan: I think the 10 is reasonable just to get a bit away from the "quantum foam" right at `::min()`, but I don't have great intuition here, if you think the 10 is bad I can remove it.

singularity-opac/photons/gray_opacity_photons.hpp
brryan: I took off the `PerNu` because unlike e.g. `EmissivityPerNu` there's no `d \nu` in the denominator here. It's still a frequency-dependent call to the absorption coefficient though. Should I go back?

singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
Yurlungur: Worth testing it, I think.

singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
Yurlungur: But this doesn't matter for that, since the lambda parameter passed into the constructor is tied to the opacity object, not the MeanOpacity. I think this function, and the `nlambda_` can go.

singularity-opac/neutrinos/non_cgs_neutrinos.hpp
Yurlungur: Ok that's fine with me.

singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
brryan: I remember why we have the `#define`... #17 I'm not sure if there's a good way around that (or a good way to have only one definition of that since you'd need an un-`undef`-ed `define` in a header file)

singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
brryan: CI upgraded

singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
brryan: OK I see yes you would only use `nlambda()` here if you wanted to see how many parameters you called the opacity with earlier, you dont actually need it to size arrays or anything during runtime. OK I'll get rid of this.

singularity-opac/neutrinos/thermal_distributions_neutrinos.hpp
brryan: As above I don't see a clear path forward for shared `EPS` at the moment due to C++14

singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
Yurlungur: oooh. Ok. I guess we can move once we move to C++17.

Yurlungur (approved):

PR #28, opened 2022-04-13
Add radiation temperature calls

brryan:
This PR adds `EnergyDensityFromTemperature`, `TemperatureFromEnergyDensity`, and `NumberDensityFromTemperature` calls to photon and neutrino thermal distributions (and loops the calls through the opacities themselves. Downstream, it turns out that having to convert between e.g. `B` and `Er` is tedious because if you aren't in cgs units you need the speed of light in code units. Also, the only way to calculate temperature from the radiation energy downstream was with a rootfind over `ThermalDistributionOfT` which is wasteful. 

I also added a simple unit test for some of these calls.


Yurlungur (approved):
LGTM

PR #29, opened 2022-04-14
Add phony thermal distribution for tests

brryan:
It can be nice for testing purposes to do radiation physics in a scale free way, which basically requires a fake thermal distribution. This PR adds a phony distribution, only to be used with a gray opacity, that just returns `1` for all the relevant quantities.


brryan:
On second thought maybe a better approach here would be a thermal distribution with constants that can be manipulated somehow, e.g. the user can set custom values of `h` (such as 1 for a scale-free solution). This would also permit radiation test problems that use modified physical constants to target specific pieces of parameter space.

singularity-opac/constants/constants.hpp
```diff
@@ -26,239 +26,186 @@
...
+  static constexpr Real mass = 1.e3;            // gram
+  static constexpr Real time = 1.;              // second
+  static constexpr Real temperature = 1.;       // Kelvin
+  static constexpr Real current = 1.e-1;        // Biot
+  static constexpr Real charge = 2.997924580e9; // Statcoulomb
+};
+
+template <typename BASE = BaseSICODATA2010,
+          typename CONVERSION = UnitConversionSIToCGS>
+struct PhysicalConstants {
```
brryan, line 100: This is a pretty drastic reimagining of the physical constants framework. Basically now it takes in a class of base constants AND a unit conversion factor. This allows for more flexibility, including fake units that still maintain some internal consistency (e.g. the radiation constant is properly derived from planck's constant). Not all unit systems can be achieved by a conversion factor (mainly for test problems that change physical constants to achieve certain solution behaviors). 

My main issue with this, I think, is that the opacity classes (opacity, thermal distribution, and mean opacity) are all now templated on the unit system. I couldn't think of an easy way to ensure consistency between all of these -- maybe we need a new container class that holds all three of these and doles out template parameters appropriately?

Yurlungur: Hmmm. Yeah I feel a little uncertain about that. A couple thoughts that may or may not address the issue:
1. Can the template parameter default to constants in CGS, so you don't have to explicitly enter a template param unless you know what you're doing?
2. Should the `Opac` object be templated on units... e.g., 
```C++
template<typename Units>
using Opac = opac_variant<GrayOpacity<Units>, TophatOpacity<Units>>
```
etc.
3. How does this interact with tabular data which *cannot* be templated on units?
4. What if the set of physical constants was an optional parameter passed into the constructor?

brryan: `1.` was my idea originally, but I would sometimes get errors saying c++14 couldn't handle that level of template deduction. I didn't really understand the problem except that it seems like nested template parameters with default arguments doesn't work in c++14. 

`2.` is interesting...

`3.` where units aren't used (e.g. tabular data and gray (scale-free) opacities) the units are just ignored. Well, the units are always used in the thermal distributions. Maybe for tabular data with specific units I should add a static assert so one must use the appropriate (CGS) units.

`4.` I'd like to keep the constants `constexpr` if at all possible

brryan: Returning to this:
`2.` and `3.` actually does only explicitly combining certain physical constants units with certain opacities in the current `using Opacity = ` not solve this problem? Someone downstream would have to manually type e.g. `SpinerOpacity<FermiDiracDistributionNoMu<3, PhysicalConstantsUnity>, PhysicalConstantsUnity>` in their code to get to an invalid state.

Yurlungur: Ah, I see what you're saying, @brryan, yes that's true. I'd stilll like to return to `1` if possible though... It would be ideal if the user didn't *have* to specify units. I don't know.

Yurlungur:
@brryan I'm willing to go with the design you have, but tests are currently not passing. Can we get tests passing and then we can merge/revisit later? 

Also does this need to be passed through the photon code path also or just skip that for now?

singularity-opac/constants/constants.hpp
brryan: So by default the user doesn't have to specify internal units and the default is cgs. Note e.g. `neutrinos::Gray opac_host(1);` in `test_gray_opacities.cpp`. This is actually consistent with default template arguments working which surprises me a bit... mean opacities still require an explicit `pc` template parameter despite the default which I don't understand.

brryan:
@Yurlungur OK CI is fixed up and `Basic` -> `ScaleFree` this should be good to go. I'd say let's punt on adding a scale-free opacity to the photon side for now, just give this a bit of time being used downstream to see if it's really what we want.

Yurlungur:
Works for me. Thanks @brryan !

Yurlungur (approved):

PR #30, opened 2022-04-15
fix assignment operator and add test

Yurlungur:
@brryan observed that the assignment operator was being implicitly deleted for singualrity opacities. There were two issues:
1. The default constructor was missing from gray opacities in photons. I added it.
2. The assignment operator was being deleted from the `BRTOpacity` class because of the constant private fields. I moved some of these into `constants.hpp` because we'll need them and made the others `static constexpr`.

I also added a test which checks that this machinery is working for the variants. To ensure everything works with C++14, the tests are now compiled with the C++14 standard.

And because it was annoying me, I deleted `test_trivial.cpp`.


brryan (approved):
LGTM, thanks for explicitly accommodating C++14

PR #31, opened 2022-05-04
Turn mean opacity into a variant

brryan:
There were some hardcoded cgs numbers in the mean opacity constructor. I fixed that and also turned mean opacity into a variant since it's annoying to use downstream otherwise.

### DIFF
--- a/singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
+++ b/singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
@@ -35,13 +35,13 @@ namespace neutrinos {
 // temperatures, and Ye
 
 template <typename pc = PhysicalConstantsCGS>
-class MeanOpacity {
+class MeanOpacityImpl {
  public:
-  MeanOpacity() = default;
+  MeanOpacityImpl() = default;
   template <typename Opacity>
-  MeanOpacity(const Opacity &opac, Real lRhoMin, Real lRhoMax, int NRho,
-              Real lTMin, Real lTMax, int NT, Real YeMin, Real YeMax, int NYe,
-              Real *lambda = nullptr) {
+  MeanOpacityImpl(const Opacity &opac, Real lRhoMin, Real lRhoMax, int NRho,
+                  Real lTMin, Real lTMax, int NT, Real YeMin, Real YeMax,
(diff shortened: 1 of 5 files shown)


brryan:
@Yurlungur I went off on my own for a bit and ended up converting the mean opacity class into a variant (only varying on the unit system of the underlying physical constants). This plays nicer downstream, for the same reason `Opacity` plays nicer as a `variant`, and I think it turned out well, but I'm not very familiar with this technique so I'm curious if you see issues with what I've done. Otherwise I think this is ready to go.

Yurlungur:
Hmm... This is an interesting idea. Let me think about it. Or maybe we could just have a quick discussion so you can walk me through your thinking. I think I like where you're going but I want to understand all the consideration.

Yurlungur:
Actually @brryan I'm confused again. I know we just talked, but there's only one mean opacity class, containing both Planck and Rosalind means. Why does that require a variant?

singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
```diff
@@ -33,13 +35,14 @@ namespace neutrinos {
 // temperatures, and Ye
 
 template <typename pc = PhysicalConstantsCGS>
-class MeanOpacity {
+class MeanOpacityImpl {
```
Yurlungur: Can you not just stick this think in the `impl` namespace again?

test/test_mean_opacities.cpp
```diff
@@ -81,8 +81,9 @@ TEST_CASE("Mean neutrino opacities", "[MeanNeutrinos]") {
     neutrinos::Gray opac_host(kappa);
     neutrinos::Opacity opac = opac_host.GetOnDevice();
 
-    neutrinos::MeanOpacity<pc> mean_opac_host(
+    neutrinos::MeanOpacityCGS mean_opac_host(
```
Yurlungur, line 84: Out of curiosity would the `MeanOpacity<CGS>` or something similar still work? Presumably yes right?

Yurlungur (approved):
Something makes me uncomfortable about doing the variant for a single class to accommodate the `constexpr` unit systems.

My preference would be to make unit systems runtime. But if that's something you'd prefer to preserve, I'm willing to go with this.

PR #34, opened 2022-10-14
Add scattering

brryan:
This PR adds scattering and mean scattering variants (gray opacity only right now) to neutrinos


singularity-opac/neutrinos/gray_s_opacity_neutrinos.hpp
```diff
@@ -0,0 +1,67 @@
...
+  GraySOpacity GetOnDevice() { return *this; }
+  PORTABLE_INLINE_FUNCTION
+  int nlambda() const noexcept { return 0; }
+  PORTABLE_INLINE_FUNCTION
+  void PrintParams() const noexcept {
+    printf("Gray scattering opacity. kappa = %g\n", kappa_);
+  }
+  inline void Finalize() noexcept {}
+
+  PORTABLE_INLINE_FUNCTION
```
Yurlungur, line 46: Since this isn't angle averaged, should it depend on mu maybe?

singularity-opac/neutrinos/mean_neutrino_s_variant.hpp
```diff
@@ -0,0 +1,104 @@
...
+#include <singularity-opac/base/radiation_types.hpp>
+#include <singularity-opac/neutrinos/neutrino_s_variant.hpp>
+#include <variant/include/mpark/variant.hpp>
+
+namespace singularity {
+namespace neutrinos {
+namespace impl {
+
+template <typename... SOpacs>
+class MeanSVariant {
```
Yurlungur, line 32: Variant here is for units vs no units?

singularity-opac/neutrinos/mean_s_opacity_neutrinos.hpp
```diff
@@ -0,0 +1,234 @@
...
+#include <singularity-opac/neutrinos/mean_neutrino_s_variant.hpp>
+#include <singularity-opac/neutrinos/thermal_distributions_neutrinos.hpp>
+
+namespace singularity {
+namespace neutrinos {
+namespace impl {
+
+#define EPS (10.0 * std::numeric_limits<Real>::min())
+
+// TODO(BRR) Note: It is assumed that lambda is constant for all densities,
```
Yurlungur, line 36: Not sure I understand this comment

singularity-opac/neutrinos/s_opac_neutrinos.hpp
```diff
@@ -0,0 +1,37 @@
...
+#include <variant/include/mpark/variant.hpp>
+
+#include <singularity-opac/neutrinos/gray_s_opacity_neutrinos.hpp>
+#include <singularity-opac/neutrinos/neutrino_s_variant.hpp>
+#include <singularity-opac/neutrinos/non_cgs_s_neutrinos.hpp>
+
+namespace singularity {
+namespace neutrinos {
+
+// TODO(JMM): Include chemical potential
```
Yurlungur, line 28: Not sure my comment applies here

Yurlungur (approved):
This all seems reasonable to me. I have some nitpicky comments below. But I also wanted to bring up a few broader comments which don't need to be addressed now:

1. Are we distinguishing between elastic and inelastic scattering here?

2. I think we should explicitly thread the angle dependence through at least somewhat here. It'll matter for inelastic scattering especially. But even for elastic, there's angle dependence. Often written in terms of my=cos(theta)

3. I am increasingly uncomfortable about how much boiler plate we're introducing for a variant that just turns units on and off. I wonder if there's some way to just put the units directly in the underlying cgs object? Or use compile-time inheritance? That can be a discussion for another time though.

I'm willing to merge now, @brryan but I'd prefer to thread through the angle dependence sooner rather than later.

Also would be interested to hear what @lroberts36 thinks as far as whether or not this infrastructure captures everything needed in the neutrino case.

singularity-opac/neutrinos/mean_neutrino_s_variant.hpp
brryan: Yes. Supposing I trade the variant over units vs no units (like this one) for a template parameter, how will that interact with parthenon-like params? For example in tasks I will want to call
```c++
auto &mean_opac = params->Get<MeanOpacity>("mean_opacity");
```
but if my type becomes `MeanOpacity<WhetherIHaveUnits>` and `WhetherIHaveUnits` is selectable at runtime when `mean_opac` is enrolled in the `Params` class, how do I pull it back out in the code later without also templating my task on `WhetherIHaveUnits` as well?

Yurlungur: I was imagining that in Phoebus, we just always have units. And non-cgs units just sets the conversion factor to 1? Is there a reason that wouldn't work? We could even do something like
```C++
typedef MeanS = MeanSWithUnits<MeanSGray>;
```
or something silly like that

singularity-opac/neutrinos/neutrino_s_variant.hpp
```diff
@@ -0,0 +1,141 @@
...
+  // rho, temp, Ye, type, nu, lambda
+  PORTABLE_INLINE_FUNCTION Real TotalScatteringCoefficient(
+      const Real rho, const Real temp, const Real Ye, const RadiationType type,
+      const Real nu, Real *lambda = nullptr) const {
+    return mpark::visit(
+        [=](const auto &s_opac) {
+          return s_opac.ScatteringCoefficient(rho, temp, Ye, type, nu, lambda);
+        },
+        s_opac_);
+  }
```
brryan, line 116: @Yurlungur How about these new functions:
```c++
Real TotalCrossSection(
    const Real rho, const Real temp, const Real Ye, const RadiationType type,
    const Real nu, Real *lambda = nullptr);
Real DifferentialCrossSection(
    const Real rho, const Real temp, const Real Ye, const RadiationType type,
    const Real nu, const Real theta, Real *lambda = nullptr);
Real TotalScatteringCoefficient(
      const Real rho, const Real temp, const Real Ye, const RadiationType type,
      const Real nu, Real *lambda = nullptr);
```

singularity-opac/neutrinos/gray_s_opacity_neutrinos.hpp
```diff
@@ -0,0 +1,71 @@
...
+#include <singularity-opac/base/opac_error.hpp>
+
+namespace singularity {
+namespace neutrinos {
+
+template <typename pc = PhysicalConstantsCGS>
+class GraySOpacity {
+ public:
+  GraySOpacity() = default;
+  GraySOpacity(const Real sigma, const Real mmw) : sigma_(sigma), mmw_(mmw) {}
```
brryan: Unlike on the absorption opacity side now we're dealing with both scattering coefficients and cross sections, so I think a reasonable solution right now is to pass mean molecular weight `mmw` to constructors generally to convert between mass density and number density. If MMW varies inside a calculation (due to Ye in NS mergers?) we may need to revisit this model.

singularity-opac/neutrinos/neutrino_s_variant.hpp
Yurlungur: that works. Is there a reason to use theta vs mu? Probably doesn't matter much, but mu might let us offload some of the work of computing transcendental functions

singularity-opac/neutrinos/gray_s_opacity_neutrinos.hpp
Yurlungur: is "molecular" weight the right concept here? More like molar weight or just average particle mass? Also it does for sure change for NS mergers, but it's incorporated already in opacity tables for the most part, which assume an EOS

brryan: "mean molecular weight" is typical in astrophysics even for this sort of thing I think, but since these indeed aren't typically molecules I switched to your `avg_particle_mass`.

OK when we add tabulated NS scattering opacities we can just leverage the data appropriately to return both `sigma` and `kappa` rather than providing an `avg_particle_mass` value to the constructor.

singularity-opac/neutrinos/neutrino_s_variant.hpp
brryan: Yeah I think `theta` will always appear inside a trigonometric function so if the code already has `mu` there is no reason to have them turn it into ` theta` before we turn it back into a `mu` or `1 - mu^2`.

singularity-opac/neutrinos/s_opac_neutrinos.hpp
brryan: Removed :+1:

singularity-opac/neutrinos/mean_s_opacity_neutrinos.hpp
brryan: We provide only one `Real *lambda` to the mean opacity constructor and these values in `lambda` are the same for every opacity call as we build up our table/integrals over density, temperature, and Ye.

singularity-opac/neutrinos/mean_neutrino_s_variant.hpp
brryan: Hm I'd have to think about this a bit more... Let me open an issue since I've already got a lot of different contexts loaded into my brain's RAM right now: #35

singularity-opac/neutrinos/gray_s_opacity_neutrinos.hpp
brryan: Fixed by the move to total vs differential cross sections

brryan:
OK @Yurlungur I think this is ready to go now if you're happy with it. Thanks for the careful review.

singularity-opac/neutrinos/neutrino_s_variant.hpp
brryan: EDIT: I did indeed make this change `theta` -> `mu`

lroberts36:
> Also would be interested to hear what @lroberts36 thinks as far as whether or not this infrastructure captures everything needed in the neutrino case.

The most general objects one would need are the scattering and absorption differential cross sections per volume for species $i$, which includes inelastic scattering (with energy exchange $q_0$)  
$$\frac{1}{V}\frac{d \sigma_{\nu_i}}{d \Omega d q_0}(E_\nu, q_0, \mu, n_b, T, Y_e).$$
In the most general case there are only single differential cross-sections for the entire medium, since the neutrons, protons, electrons, etc. may be interacting with one another. In practice it may make sense to separate out electrons for scattering since $q_0 \sim E_\nu$ in that case. 

For scattering, it looks like the current infrastructure is setup to return 
$$\frac{d \sigma_{\nu_i}}{d \Omega}(E_\nu, \mu, n_b, T, Y_e) = \int_{-\infty}^{E_\nu} dq_0 \frac{d \sigma_{\nu_i}}{d \Omega d q_0},$$
with the energy integration being somewhat reasonable for nucleons since the group width is likely to be larger than the energy transfer to/from the nucleons. That being said, the nucleon scattering cross section are nearly isotropic and are well approximated by 
$$\frac{d \sigma_{\nu_i}}{d \Omega} = \frac{\sigma_t(E_\nu, n_b, T, Y_e)}{4 \pi} (1 + \delta \mu),$$
where $\delta$ is a constant made up of various weak interaction coupling constants. As a result, it might be more efficient and accurate to store and return angular moments of both the inelastic and elastic scattering cross-sections
$$\sigma_{\nu_i, \ell} = 2 \pi \int_{-1}^{1} d\mu P_\ell(\mu) \frac{d \sigma_{\nu_i}}{d \Omega}.$$
Inelastic scattering from electrons is also well approximated by the first three moments (and explicit expressions for the moments are available in the literature). I guess this would really be an under the hood change, since we still probably want methods that get the cross-section as a function of $\mu$, independent of how they are stored. 

It seems like including inelastic scattering will require some generalization of the infrastructure. 

@Yurlungur

Yurlungur:
Thanks @lroberts36 . In nubhlight, I used exactly the elastic scattering off of nucleons... Plus corrections for things like non-sphericity, shielding etc, which made it a bit nastier. Does inelastic scattering require a change to the API do you think? Or can it be hidden under the hood? Sounds like perhaps if we can use a moment approximation, the under-the-hood approach would work?

lroberts36:
@Yurlungur I think inelastic scattering requires a extension of the API to include a method that takes $E_\nu$, $q_0$, $\mu$, and ( $n_b$, $T$, $Y_e$) and returns the differential cross section.

Yurlungur:
Do you think that should be a new function or should the current functions be extended? I lean towards a new function specifically for inelastic, but I could be convinced the other way

lroberts36:
I think it should be a separate function.

brryan:
This all seems reasonable to me

PR #36, opened 2022-10-25
Robustify mean opacity calls

brryan:
Sometimes we want to evaluate mean opacities even when the opacity is zero. This code robustifies the evaluation of the Rosseland mean, which was dividing by zero and producing NANs.


Yurlungur (approved):
:+1:

PR #37, opened 2022-10-25
Thomson and Bremss opacities

brryan:
I add photon variants for Thomson scattering opacity and electron-proton bremsstrahlung opacity. 

Along the way I add a mean scattering class to photons analogous to the recently added mean scattering class for neutrinos #34

### DIFF
--- a/singularity-opac/photons/epbremsstrahlung_opacity_photons.hpp
+++ b/singularity-opac/photons/epbremsstrahlung_opacity_photons.hpp
@@ -27,6 +27,8 @@
 namespace singularity {
 namespace photons {
 
+// Expression for specific emissivity from Rybicki & Lightman 1979
+
 template <typename pc = PhysicalConstantsCGS>
 class EPBremsstrahlungOpacity {
  public:
--- a/singularity-opac/photons/epbremsstrahlung_opacity_photons.hpp
+++ b/singularity-opac/photons/epbremsstrahlung_opacity_photons.hpp
@@ -180,9 +180,9 @@ class EPBremsstrahlungOpacity {
       (pc::mp + pc::me) /
       (2. * pc::mp); // Neutral fully ionized electron-proton gas
   static constexpr Real gff_ = 1.2;
-  static constexpr Real prefac_ = 8. * std::pow(pc::qe, 6) /
-                                  (3. * std::pow(pc::me * pc::c * pc::c, 2)) *
-                                  std::sqrt(2. * M_PI / 3.);
+  Real prefac_ = 8. * std::pow(pc::qe, 6) /
+                 (3. * std::pow(pc::me * pc::c * pc::c, 2)) *
+                 std::sqrt(2. * M_PI / 3.);
 };
 
 } // namespace photons
--- a/singularity-opac/photons/epbremsstrahlung_opacity_photons.hpp
+++ b/singularity-opac/photons/epbremsstrahlung_opacity_photons.hpp
@@ -1,0 +1,189 @@
+// ======================================================================
+// © 2022. Triad National Security, LLC. All rights reserved.  This
+// program was produced under U.S. Government contract
+// 89233218CNA000001 for Los Alamos National Laboratory (LANL), which
+// is operated by Triad National Security, LLC for the U.S.
+// Department of Energy/National Nuclear Security Administration. All
+// rights in the program are reserved by Triad National Security, LLC,
+// and the U.S. Department of Energy/National Nuclear Security
+// Administration. The Government is granted for itself and others
+// acting on its behalf a nonexclusive, paid-up, irrevocable worldwide
+// license in this material to reproduce, prepare derivative works,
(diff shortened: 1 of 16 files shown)


singularity-opac/photons/epbremsstrahlung_opacity_photons.hpp
```diff
@@ -0,0 +1,191 @@
...
+    const Real ni = ne;
+    return 4. * M_PI * pc::kb * temp / pc::h * prefac_ / std::sqrt(thetaE) *
+           ne * ni * gff_;
+  }
+
+  PORTABLE_INLINE_FUNCTION
+  Real NumberEmissivity(const Real rho, const Real temp,
+                        Real *lambda = nullptr) const {
+    // Infrared catastrophe
+    return std::numeric_limits<Real>::infinity();
```
brryan, line 133: \int d\Omega d\nu (1 / (h \nu)) exp^{-h \nu / (k T)}

Yurlungur: lol :+1:

singularity-opac/photons/epbremsstrahlung_opacity_photons.hpp
```diff
@@ -0,0 +1,191 @@
...
+
+ private:
+  Real mass_ion_;
+  PlanckDistribution<pc> dist_;
+  static constexpr Real mmw_ =
+      (pc::mp + pc::me) /
+      (2. * pc::mp); // Neutral fully ionized electron-proton gas
+  static constexpr Real gff_ = 1.2;
+  static constexpr Real prefac_ = 8. * std::pow(pc::qe, 6) /
+                                  (3. * std::pow(pc::me * pc::c * pc::c, 2)) *
```
Yurlungur: didn't know `std::pow` was constexpr... nice.

singularity-opac/photons/s_opac_photons.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+#include <singularity-opac/photons/gray_s_opacity_photons.hpp>
+#include <singularity-opac/photons/non_cgs_s_photons.hpp>
+#include <singularity-opac/photons/photon_s_variant.hpp>
+#include <singularity-opac/photons/thomson_s_opacity_photons.hpp>
+
+namespace singularity {
+namespace photons {
+
+// TODO(JMM): Include chemical potential
```
Yurlungur: probably not relevant for photons

Yurlungur (approved):
My reservations about the amount of boiler plate continue to grow, but otherwise looks good. 

I wonder if we could re-use some of the mean opacity stuff accross photons/neutrinos, because it's ALMOST identical, modulo Ye.

singularity-opac/photons/s_opac_photons.hpp
brryan: Right thanks, fixed

singularity-opac/photons/epbremsstrahlung_opacity_photons.hpp
brryan: Actually this is a good point, it looks like `std::pow` being `constexpr` is an explicitly standards-breaking feature of gcc. I should go ahead and change this now before it causes problems later.

brryan: OK just removed the `static constexpr` decorators. In the future if we do want expressions like this to be `constexpr` we could add a dependency to [Sprout](https://github.com/bolero-MURAKAMI/Sprout)

brryan:
Re: boilerplate, I agree, though I'm not immediately sure what the right thing to do is, and also I at least right now don't anticipate creating any more variants :)

singularity-opac/photons/epbremsstrahlung_opacity_photons.hpp
Yurlungur: these are also integer powers, so we could just write them out, or even implement our own constexpr pow. Something along the lines of the code here:
https://stackoverflow.com/questions/17719674/c11-fast-constexpr-integer-powers

Yurlungur: From stack overflow:
```C++
#include <cstdint>

constexpr int64_t ipow(int64_t base, int exp, int64_t result = 1) {
  return exp < 1 ? result : ipow(base*base, exp/2, (exp % 2) ? result*base : result);
}

int64_t foo(int64_t base, int exp) {
  return ipow(base, exp);
}

```

Yurlungur: Something like... (checked in godbolt):
```C++
constexpr Real mypow(Real x, std::size_t p) {
    if (p == 0) return 1;
    if (p == 1) return x;
    if ((p % 2) == 0) return mypow(x, p/2)*mypow(x, p/2);
    return mypow(x, p - 1)*x;
}
```

PR #38, opened 2022-10-26
Additional Rosseland opacity fixes

brryan:
I was still running into `infinity` values from the Rosseland mean opacity so I modified the code to only compute the integrand if `alpha > singularity_opac::robust::SMALL()` for that support point. 

Along the way I cleaned up the code for the trapezoidal rule, reducing redundant opacity evaluations. 

Mean opacities can also now be evaluated either with a default per-rho, T, Ye frequency range, or with a common frequency range specified by an overloaded constructor

### DIFF
--- a/singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
+++ b/singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
@@ -46,8 +46,9 @@ class MeanOpacity {
               const int NRho, const Real lTMin, const Real lTMax, const int NT,
               const Real YeMin, const Real YeMax, const int NYe,
               Real *lambda = nullptr) {
-    MeanOpacityImpl<Opacity, true>(opac, lRhoMin, lRhoMax, NRho, lTMin, lTMax, NT, YeMin, YeMax,
-                NYe, -1., -1., 100, lambda);
+    MeanOpacityImpl<Opacity, true>(opac, lRhoMin, lRhoMax, NRho, lTMin, lTMax,
+                                   NT, YeMin, YeMax, NYe, -1., -1., 100,
+                                   lambda);
   }
 
   template <typename Opacity>
@@ -55,15 +56,17 @@ class MeanOpacity {
               const int NRho, const Real lTMin, const Real lTMax, const int NT,
               const Real YeMin, const Real YeMax, const int NYe, Real lNuMin,
               Real lNuMax, const int NNu, Real *lambda = nullptr) {
-    MeanOpacityImpl<Opacity, false>(opac, lRhoMin, lRhoMax, NRho, lTMin, lTMax, NT, YeMin, YeMax,
-                NYe, lNuMin, lNuMax, NNu, lambda);
+    MeanOpacityImpl<Opacity, false>(opac, lRhoMin, lRhoMax, NRho, lTMin, lTMax,
+                                    NT, YeMin, YeMax, NYe, lNuMin, lNuMax, NNu,
+                                    lambda);
   }
 
   template <typename Opacity, bool AUTOFREQ>
-  MeanOpacityImpl(const Opacity &opac, const Real lRhoMin, const Real lRhoMax,
-              const int NRho, const Real lTMin, const Real lTMax, const int NT,
-              const Real YeMin, const Real YeMax, const int NYe, Real lNuMin,
-              Real lNuMax, const int NNu, Real *lambda = nullptr) {
+  void MeanOpacityImpl(const Opacity &opac, const Real lRhoMin,
+                       const Real lRhoMax, const int NRho, const Real lTMin,
+                       const Real lTMax, const int NT, const Real YeMin,
+                       const Real YeMax, const int NYe, Real lNuMin,
+                       Real lNuMax, const int NNu, Real *lambda = nullptr) {
     lkappaPlanck_.resize(NRho, NT, NYe, NEUTRINO_NTYPES);
     // index 0 is the species and is not interpolatable
     lkappaPlanck_.setRange(1, YeMin, YeMax, NYe);
@@ -86,7 +89,8 @@ class MeanOpacity {
             Real kappaPlanckDenom = 0.;
             Real kappaRosselandNum = 0.;
             Real kappaRosselandDenom = 0.;
-            // Choose default temperature-specific frequency grid if frequency grid not specified
+            // Choose default temperature-specific frequency grid if frequency
+            // grid not specified
             if (AUTOFREQ) {
               lNuMin = toLog_(1.e-3 * pc::kb * fromLog_(lTMin) / pc::h);
               lNuMax = toLog_(1.e3 * pc::kb * fromLog_(lTMax) / pc::h);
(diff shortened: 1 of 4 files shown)


singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
```diff
@@ -46,6 +46,24 @@ class MeanOpacity {
...
+              Real lNuMax, const int NNu, Real *lambda = nullptr) {
+    MeanOpacityImpl<Opacity, false>(opac, lRhoMin, lRhoMax, NRho, lTMin, lTMax, NT, YeMin, YeMax,
+                NYe, lNuMin, lNuMax, NNu, lambda);
+  }
+
+  template <typename Opacity, bool AUTOFREQ>
+  MeanOpacityImpl(const Opacity &opac, const Real lRhoMin, const Real lRhoMax,
+              const int NRho, const Real lTMin, const Real lTMax, const int NT,
+              const Real YeMin, const Real YeMax, const int NYe, Real lNuMin,
+              Real lNuMax, const int NNu, Real *lambda = nullptr) {
```
Yurlungur: Does this actually work? I think your constructor **has** to be named the same as the class. You could, however, use tag dispatch or something?

singularity-opac/neutrinos/mean_s_opacity_neutrinos.hpp
```diff
@@ -46,6 +46,24 @@ class MeanSOpacity {
...
+               Real lNuMax, const int NNu, Real *lambda = nullptr) {
+    MeanSOpacityImpl<SOpacity, false>(s_opac, lRhoMin, lRhoMax, NRho, lTMin, lTMax, NT, YeMin, YeMax,
+                 NYe, lNuMin, lNuMax, 100, lambda);
+  }
+
+  template <typename SOpacity, bool AUTOFREQ>
+  MeanSOpacityImpl(const SOpacity &s_opac, const Real lRhoMin, const Real lRhoMax,
+               const int NRho, const Real lTMin, const Real lTMax, const int NT,
+               const Real YeMin, const Real YeMax, const int NYe, Real lNuMin,
+               Real lNuMax, const int NNu, Real *lambda = nullptr) {
```
Yurlungur: See above concern about impl constructor

singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
Yurlungur: Does this actually work? I think the constructor has to be named exactly the same as the class. Maybe you need to go to tag dispatch or something?

singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
```diff
@@ -68,68 +86,45 @@ class MeanOpacity {
...
-            lkappaPlanck_(iRho, iT, iYe, idx) = lkappaPlanck;
-            lkappaRosseland_(iRho, iT, iYe, idx) = lkappaRosseland;
+            Real kappaPlanck = singularity_opac::robust::ratio(
+                kappaPlanckNum, kappaPlanckDenom);
+            Real kappaRosseland =
+                kappaPlanck > singularity_opac::robust::SMALL()
+                    ? singularity_opac::robust::ratio(kappaRosselandDenom,
+                                                      kappaRosselandNum)
+                    : 0.;
+            printf("kappas: %e %e\n", kappaPlanck, kappaRosseland);
```
Yurlungur: Should this printf still be here? If so might be good to make it more descriptive

singularity-opac/photons/mean_opacity_photons.hpp
```diff
@@ -45,6 +45,23 @@ class MeanOpacity {
...
+              Real lNuMin, Real lNuMax, const int NNu, Real *lambda = nullptr) {
+    MeanOpacityImpl<Opacity, false>(opac, lRhoMin, lRhoMax, NRho, lTMin, lTMax,
+                                    NT, lNuMin, lNuMax, NNu, lambda);
+  }
+
+  template <typename Opacity, bool AUTOFREQ>
+  MeanOpacityImpl(const Opacity &opac, const Real lRhoMin, const Real lRhoMax,
+                  const int NRho, const Real lTMin, const Real lTMax,
+                  const int NT, Real lNuMin, Real lNuMax, const int NNu,
+                  Real *lambda = nullptr) {
```
Yurlungur: See above concern about `impl`

singularity-opac/photons/mean_s_opacity_photons.hpp
```diff
@@ -46,6 +46,24 @@ class MeanSOpacity {
...
+               Real *lambda = nullptr) {
+    MeanSOpacityImpl<SOpacity, false>(s_opac, lRhoMin, lRhoMax, NRho, lTMin,
+                                      lTMax, NT, lNuMin, lNuMax, NNu, lambda);
+  }
+
+  template <typename SOpacity, bool AUTOFREQ>
+  MeanSOpacityImpl(const SOpacity &s_opac, const Real lRhoMin,
+                   const Real lRhoMax, const int NRho, const Real lTMin,
+                   const Real lTMax, const int NT, Real lNuMin, Real lNuMax,
+                   const int NNu, Real *lambda = nullptr) {
```
Yurlungur: see above question about `impl` constructor

Yurlungur (changes requested):
Overall nice cleanup. :+1: Looks like tests are failing though and I suspect I know why. See below:

singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
brryan: Oh yeah you're right, I meant `MeanOpacityImpl` to be a `void` function. I changed that here and elsewhere and the tests pass without memory failures. I have no idea what the compiler was doing before that change...

singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
brryan: Oops yeah that was garbage, gone

brryan:
OK this should be good now @Yurlungur

singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
```diff
@@ -46,6 +46,27 @@ class MeanOpacity {
...
+              const int NRho, const Real lTMin, const Real lTMax, const int NT,
+              const Real YeMin, const Real YeMax, const int NYe, Real lNuMin,
+              Real lNuMax, const int NNu, Real *lambda = nullptr) {
+    MeanOpacityImpl<Opacity, false>(opac, lRhoMin, lRhoMax, NRho, lTMin, lTMax,
+                                    NT, YeMin, YeMax, NYe, lNuMin, lNuMax, NNu,
+                                    lambda);
+  }
+
+  template <typename Opacity, bool AUTOFREQ>
+  void MeanOpacityImpl(const Opacity &opac, const Real lRhoMin,
```
Yurlungur: As a void function should `MeanOpacityImpl` have public scope? Also if private scope, should have trailing underscore

Yurlungur (changes requested):

singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
brryan: Ah yes these four `void` functions should be `private` and have a trailing underscore -- done, thanks

Yurlungur (approved):
:+1:

PR #40, opened 2022-10-28
Fix unset unit conversions

brryan:
Somehow, I left some unit conversions in the non-cgs photons class unset. This PR fixes that.


singularity-opac/neutrinos/s_opac_neutrinos.hpp
```diff
@@ -25,7 +25,6 @@
 namespace singularity {
 namespace neutrinos {
 
-// TODO(JMM): Include chemical potential
```
brryan, line 28: I removed this previously, but it didnt get committed into an earlier PR, oops (too many computers)

Yurlungur (approved):

PR #41, opened 2022-10-31
Fix default constructors

brryan:
I wasn't able to properly create default versions of all the opacity variants due to some confusion probably due just to `noexcept`, but this PR also standardizes the lack of GPU decorators for default constructors.

### DIFF
--- a/singularity-opac/photons/epbremsstrahlung_opacity_photons.hpp
+++ b/singularity-opac/photons/epbremsstrahlung_opacity_photons.hpp
@@ -177,8 +177,7 @@ class EPBremsstrahlungOpacity {
   Real mass_ion_;
   PlanckDistribution<pc> dist_;
   static constexpr Real mmw_ =
-      (pc::mp + pc::me) /
-      (2. * pc::mp); // Neutral fully ionized electron-proton gas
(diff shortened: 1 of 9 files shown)


singularity-opac/photons/epbremsstrahlung_opacity_photons.hpp
```diff
@@ -177,8 +177,7 @@ class EPBremsstrahlungOpacity {
   Real mass_ion_;
   PlanckDistribution<pc> dist_;
   static constexpr Real mmw_ =
-      (pc::mp + pc::me) /
-      (2. * pc::mp); // Neutral fully ionized electron-proton gas
+      (pc::mp + pc::me) / 2.; // Neutral fully ionized electron-proton gas (g)
```
Yurlungur, line 180: double checking---is this change right? Going from unitless to grams?

Yurlungur (approved):

singularity-opac/photons/epbremsstrahlung_opacity_photons.hpp
brryan: yes, this makes more sense because now `ne = ni  = rho / mmw`

PR #46, opened 2023-07-22
CMake updates

mauneyc-LANL:
Lots has changed in how we build, so I've made some fixes. This isn't a complete redo of the build, which we should do but this shouldn't be too much work.

- Removed `::flags` and `::hdf5` flags
- explicit target names `singularity-opac`
- CMake formatting
- `ports-of-call` submodule update (do we need `spiner` for this project?)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,11 +15,13 @@ jobs:
         - name: Checkout code
           uses: actions/checkout@v2
           with:
-            submodules: true
+            submodules: recursive
         - name: Set system to non-interactive mode
           run: export DEBIAN_FRONTEND=noninteractive
         - name: install dependencies
-          run: sudo apt-get install -y --force-yes -qq build-essential libhdf5-serial-dev
+          run: |
+	    sudo apt-get update -y -qq
+	    sudo apt-get install -y --allow-downgrades --allow-remove-essential --allow-change-held-packages -qq build-essential libhdf5-serial-dev
         - name: build and run tests
           run: |
             mkdir -p bin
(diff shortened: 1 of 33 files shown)


mauneyc-LANL:
The CI probably needs some attending too:

```bash
Run sudo apt-get install -y --force-yes -qq build-essential libhdf5-serial-dev
W: --force-yes is deprecated, use one of the options starting with --allow instead.
E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/c/curl/libcurl4-openssl-dev_7.81.0-1ubuntu1.10_amd64.deb  404  Not Found [IP: 52.250.76.244 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
Error: Process completed with exit code 100.
```

Yurlungur (approved):
Thanks for taking a look at this, @mauneyc-LANL ! Yes---we do need spiner for this project. I'll look at what's going on with the CI

Yurlungur:
@mauneyc-LANL I fixed the CI tests are now failing for other reasons. Looks like missing (or changed?) include paths. Did you want to mess with it? Or should I take a look?

mauneyc-LANL:
@jonahm-LANL Good bit of changes needed. Not exactly sure what the gray opaq test is doing yet, but everything else is moving.

There's still lots of work that needs to be done to get this in a state for a stand-alone distribution. I'm assuming that the immediate need is for a submodule, so some shortcuts when sorting out dependencies/definitions.

I would further like to discourage all devs from copy-pasting code.

PR #47, opened 2023-12-08
Add template args to `DataBox`

brryan:
We started getting compiler errors from `DataBox` due to lack of template argument. I added `<Real>` to the neutrino and photon `DataBox`es used by the mean opacity functions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-# © 2021. Triad National Security, LLC. All rights reserved.  This
+# © 2023. Triad National Security, LLC. All rights reserved.  This
 # program was produced under U.S. Government contract
 # 89233218CNA000001 for Los Alamos National Laboratory (LANL), which
 # is operated by Triad National Security, LLC for the U.S.  Department
@@ -12,6 +12,7 @@
 # publicly, and to permit others to do so.
 
 cmake_minimum_required(VERSION 3.14)
+set(CMAKE_CXX_STANDARD 17 CACHE BOOL "")
 
 project("singularity-opac"
       VERSION 0.0.1
@@ -53,10 +54,20 @@ target_include_directories(singularity-opac::flags
                            INTERFACE
                            $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>)
 
-include (SetupDeps)
 include (SetupOptions)
+include (SetupDeps)
 include (SetupCompilers)
 include (SetupFlags)
+
+if (SINGULARITY_USE_HDF5)
+  message(STATUS "HDF5 enabled. Requesting it in spiner")
+  set(SPINER_USE_HDF ON CACHE BOOL "")
+else()
+  message(STATUS "HDF5 not enabled. Turning it off in spiner")
+  set(SPINER_USE_HDF OFF CACHE BOOL "")
+endif()
+add_subdirectory(utils/spiner)
+target_link_libraries(singularity-opac::flags INTERFACE spiner)
 
 include(GNUInstallDirs)
 include(CTest)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,6 @@
 # publicly, and to permit others to do so.
 
 cmake_minimum_required(VERSION 3.14)
-set(CMAKE_CXX_STANDARD 17 CACHE BOOL "")
 
 project("singularity-opac"
       VERSION 0.0.1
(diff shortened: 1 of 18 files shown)


brryan:
I'm not sure what's  going on with the CI here `E: Failed to fetch mirror+file:/etc/apt/apt-mirrors.txt/pool/main/c/curl/libcurl4-openssl-dev_7.81.0-1ubuntu1.14_amd64.deb  404  Not Found [IP: 40.81.13.82 80]` @Yurlungur @mauneyc-LANL

Yurlungur:
> I'm not sure what's going on with the CI here `E: Failed to fetch mirror+file:/etc/apt/apt-mirrors.txt/pool/main/c/curl/libcurl4-openssl-dev_7.81.0-1ubuntu1.14_amd64.deb 404 Not Found [IP: 40.81.13.82 80]` @Yurlungur @mauneyc-LANL

I bet the VM changed. Let me see if I can fix it

Yurlungur:
@brryan I fixed the CI--needed to update the workflows clone call. Some bits have rotted with the tests, though, in particular with ports-of-call.

Yurlungur (approved):
LGTM

brryan:
Thanks @Yurlungur! Can I go ahead and merge even with the `ports-of-call` issue in CI, or are you still looking at that?

Yurlungur:
I haven't figure it out. Something screwy with the cmake. I may need @mauneyc-LANL 's help

cmauney:
Can you clarify is this `spiner` or `ports-of-call`?

cmauney:
Looking over `cmake/SetupDeps.cmake`, I don't see the `spiner` target added to `singularity-opaq::flags`, and the tests don't bring in `spiner::spiner`, which is probably why the tests are giving the include errors.

Yurlungur:
> Can you clarify is this `spiner` or `ports-of-call`?

There were a number of issues. With an update to `spiner`, `singularity-opac` needed a copy of `ports-of-call` again. But also HDF5 is being a problem. Whether or not spiner should use it is not being appropriately communicated between `singularity-opac` and `spiner`.

Yurlungur:
OK I think I fixed it. @brryan @cmauney @mauneyc-LANL please take a look.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-# © 2021. Triad National Security, LLC. All rights reserved.  This
+# © 2023. Triad National Security, LLC. All rights reserved.  This
```
brryan: 2021-2023?

cmake/SetupDeps.cmake
```diff
@@ -64,9 +66,22 @@ if (HDF5_FOUND)
       else()
 	message("MPI::MPI_CXX provided by parent package")
       endif()
+    else()
+      message(status "HDF5 is not parallel")
```
brryan, line 70: `message(STATUS`?

singularity-opac/photons/mean_opacity_photons.hpp
```diff
@@ -186,8 +186,8 @@ class MeanOpacity {
   PORTABLE_INLINE_FUNCTION Real fromLog_(const Real lx) const {
     return std::pow(10., lx);
   }
-  Spiner::DataBox lkappaPlanck_;
-  Spiner::DataBox lkappaRosseland_;
+  Spiner::DataBox<Real> lkappaPlanck_;
```
brryan: Thanks for fixing the `mean_s_opacity` functions as well -- should we move these original changes to the `using Databox = ` pattern you have elsewhere for consistency? I can do this if you want to hand the PR back to me

brryan (commented):
LGTM

cmauney:
I still haven't wrapped my head around the original issue, but I don't see any problems with changes.If it works g2g

cmake/SetupDeps.cmake
cmauney: case-senstivity is weird in CMake but I think in this case it's insensitive

brryan: I briefly googled this out of curiosity and the first example was `message(status` of all things and apparently it actually does matter in this case: https://stackoverflow.com/a/35034125

CMakeLists.txt
Yurlungur: :+1:

cmake/SetupDeps.cmake
Yurlungur: It looks like that stack overflow is about the case of the thing the message is about, not the STATUS flag? The rest of the code is using lower-case, which is why I kept doing it

singularity-opac/photons/mean_opacity_photons.hpp
Yurlungur: Sure, that sounds good. Thanks!

Yurlungur:
@brryan if you wanna change the `mean_opacity` to the `using` pattern, go for it. Or just merge at will, either way is good with me. :+1:

PR #48, opened 2024-01-18
Fix return value

brryan:
A `DataIndexer`-based photon absorption call was marked as a `Real` rather than `void` function even though it should not have a return value. Compiler warning downstream.


Yurlungur (approved):
Thanks for the fix!

PR #50, opened 2024-03-15
Use https for ports of call submodule

AstroBarker:
The `ports-of-call` submodule was listed in `.gitmodules` as using ssh instead of https. It has been changed to https.


Yurlungur (approved):
Thanks for this fix!

PR #51, opened 2024-09-25
Power law opacity

brryan:
Generic power law of density and temperature for the opacity.

### DIFF
--- a/cmake/SetupDeps.cmake
+++ b/cmake/SetupDeps.cmake
@@ -93,3 +93,5 @@ cmake_dependent_option(SINGULARITY_USE_MPI
 if (NOT TARGET Catch2::Catch2)
   find_package(Catch2 QUIET)
 endif()
+
+message(STATUS "NOW Compile Definitions for ${PROJECT_NAME}: ${defs}")
--- a/cmake/SetupDeps.cmake
+++ b/cmake/SetupDeps.cmake
@@ -21,10 +21,13 @@ target_link_libraries(singularity-opac::flags INTERFACE PortsofCall::PortsofCall
 # Setup Kokkos
 # - provides Kokkos::kokkos
 #=======================================
-if (NOT TARGET Kokkos::kokkos)
-  find_package(Kokkos QUIET)
-else()
-  message(status "Kokkos::kokkos provided by parent package")
+if (SINGULARITY_USE_KOKKOS)
+  if (NOT TARGET Kokkos::kokkos)
+    add_subdirectory(utils/kokkos)
+    find_package(Kokkos QUIET)
+  else()
+    message(status "Kokkos::kokkos provided by parent package")
+  endif()
 endif()
 
 #=======================================
--- a/cmake/SetupDeps.cmake
+++ b/cmake/SetupDeps.cmake
@@ -11,22 +11,30 @@ else()
 endif()
 
 #=======================================
+# Setup Kokkos
+# - provides Kokkos::kokkos
+#=======================================
+if (SINGULARITY_USE_KOKKOS)
+  if (NOT TARGET Kokkos::kokkos)
+    message(status "Kokkos::kokkos must be found")
+    if (SINGULARITY_KOKKOS_IN_TREE)
+      message(status "Using in-tree Kokkos")
+      add_subdirectory(utils/kokkos)
+    else()
+      message(status "Using system Kokkos if available")
+      find_package(Kokkos REQUIRED)
+    endif()
+  else()
+    message(status "Kokkos::kokkos provided by parent package")
+  endif()
+endif()
+
+#=======================================
 # Setup ports of call
 # - provides PortsofCall::PortsofCall
 #=======================================
 find_package(PortsofCall REQUIRED)
 target_link_libraries(singularity-opac::flags INTERFACE PortsofCall::PortsofCall)
-
-#=======================================
-# Setup Kokkos
-# - provides Kokkos::kokkos
-#=======================================
-if (NOT TARGET Kokkos::kokkos)
-  add_subdirectory(utils/kokkos)
-  find_package(Kokkos QUIET)
-else()
-  message(status "Kokkos::kokkos provided by parent package")
-endif()
 
 #=======================================
 # Find HDF5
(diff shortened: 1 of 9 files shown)


singularity-opac/photons/powerlaw_opacity_photons.hpp
```diff
@@ -0,0 +1,181 @@
...
+      coeffs[i] =
+          AngleAveragedAbsorptionCoefficient(rho, temp, nu_bins[i], lambda);
+    }
+  }
+
+  PORTABLE_INLINE_FUNCTION
+  Real EmissivityPerNuOmega(const Real rho, const Real temp, const Real nu,
+                            Real *lambda = nullptr) const {
+    Real Bnu = dist_.ThermalDistributionOfTNu(temp, nu, lambda);
+    return rho * (kappa0_ * std::pow(rho, A_) * std::pow(temp, B_) * Bnu;
```
pdmullen: Missing parenthesis?

singularity-opac/photons/powerlaw_opacity_photons.hpp
```diff
@@ -0,0 +1,181 @@
...
+#include <singularity-opac/photons/thermal_distributions_photons.hpp>
+
+namespace singularity {
+namespace photons {
+
+template <typename pc = PhysicalConstantsCGS>
+class PowerLawOpacity {
+ public:
+  PowerLawOpacity() = default;
+  PowerLawOpacity(const Real kappa0, const Real A, const Real B)
```
pdmullen: Below you use B in several places for the Planck function.  Can we reserve a different variable name for the power law exponent?

pdmullen (approved):
Thanks for adding this!

test/test_powerlaw_opacities.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+    opac.Finalize();
+  }
+}
+
+TEST_CASE("Tophat Opacities", "[TopHat]") {
+  WHEN("We initialize a tophat neutrino opacity") {
+    neutrinos::Opacity opac = neutrinos::Tophat(1, 1e-2, 1e2);
+  }
+}
```
Yurlungur: I assume this is leftover from a copy-paste?

Yurlungur (approved):
Thanks for adding this! This will be important for us in riot too.

singularity-opac/photons/powerlaw_opacity_photons.hpp
brryan: `A -> rho_exp`, `B -> temp_exp`

test/test_powerlaw_opacities.cpp
brryan: Yep, removed :+1:

cmake/SetupDeps.cmake
```diff
@@ -22,6 +22,7 @@ target_link_libraries(singularity-opac::flags INTERFACE PortsofCall::PortsofCall
 # - provides Kokkos::kokkos
 #=======================================
 if (NOT TARGET Kokkos::kokkos)
+  add_subdirectory(utils/kokkos)
```
brryan: I needed to make this change to compile with kokkos

cmake/SetupFlags.cmake
```diff
@@ -72,6 +72,7 @@ INTERFACE
     $<${with_fmath}:
       SINGULARITY_USE_FMATH
     >
+    PORTABILITY_STRATEGY_NONE
```
brryan: This seems necessary... but `PORTABILITY_STRATEGY_KOKKOS` is *always* provided to `make` as a flag for any set of options I tried.

cmake/SetupOptions.cmake
```diff
@@ -51,3 +51,9 @@ if(SINGULARITY_USE_CUDA)
     message(FATAL_ERROR "Cuda without kokkos is not currently supported")
   endif()
 endif()
+
+if(NOT SINGULARITY_USE_KOKKOS)
+  message(FATAL_ERROR
+    "For unknown reasons PORTABILITY_STRATEGY_KOKKOS will always be provided to ports-of-call. "
+    "For now you must use Kokkos with singularity-opac.")
+endif()
```
brryan: I'm stumped by this one right now... At the end of my cmake I can't detect that `PORTABILITY_STRATEGY_KOKKOS` is set anywhere, but during `make` there's always a `-DPORTABILITY_STRATEGY_KOKKOS` which breaks compilation when `PORTABILITY_STRATEGY_NONE` is used. I'm inclined to punt on this for the purposes of this PR.

Yurlungur: I am going to try and look at this this afternoon. If I can figure it out by COB we might want to include my fix. Otherwise yeah we can punt.

brryan: Interesting data point is that the CI apparently can build the code just fine with `PORTABILITY_STRATEGY_NONE`. It's possible this is an error on my platform somehow. But it would be nice to understand why this is occurring.

cmake/SetupFlags.cmake
Yurlungur: I don't think this should be necessary. very weird.

cmake/SetupOptions.cmake
Yurlungur: I believe I fixed it. Can build on my laptop with and without kokkos. There were a few little things:
- There was an order error where options were being set AFTER dependencies. 
- SINGULARITY_KOKKOS_IN_TREE did nothing. I re-enabled it.
- If you want to disable cmake finding your pre-installed libraries, set
```C++
  set(CMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY ON CACHE BOOL "" FORCE)
  set(CMAKE_FIND_PACKAGE_NO_SYSTEM_PACKAGE_REGISTRY ON CACHE BOOL "" FORCE)
```
which are now enabled when you build with `SINGULARITY_KOKKOS_IN_TREE=ON`.

Yurlungur:
@brryan I think the build system should be fixed now... Can you give it a try?

Yurlungur:
And if it works, merge at will :)

cmake/SetupFlags.cmake
brryan: Yeah I think this is unnecessary given the default behavior of `ports-of-call`, I'll remove it

cmake/SetupOptions.cmake
brryan: Thanks @Yurlungur! This does indeed fix compilation for me.

PR #54, opened 2024-10-29
Add ASCII opacity parser to mean_opacity_photons class.

RyanWollaeger:
+ Add ASCII grey data parser to MeanOpacity class in `mean_opacity_photons.hpp`
+ Combine Rosseland and Planck opacity into one `Databox`, with a 3rd discrete index valued:
  - 0 = Rosseland,
  - 1 = Planck.
+ Add python script for generating power-law opacity ASCII file, and example file for unit test.
+ Add unit test that parses toy opacity file (`kap_plaw.txt`).


RyanWollaeger:
@brryan @jonahm-LANL - can't seem to add reviewers - anyway here is a preliminary changeset - thanks for the help!

singularity-opac/photons/zhu_data/opacitysolar09dustq3p5amax0p1new.txt
Yurlungur, line 1: not a huge fan of storing an ascii data file in the repo---but if this is the best solution we can work with it. It might be better to take a file name as input and distribute this file in a known place or have a script to fetch it. Same goes for the hdf5 binary. We could, if you like distribute them as artifacts in a data release via github's release mechanism. We do something similar for gold file data in singularity-eos.

singularity-opac/photons/zhu_grey_table_opacity_photons.hpp
```diff
@@ -0,0 +1,309 @@
...
+// DataBox use is based on, e.g., photon/mean_s_opacity_photons.hpp
+template <typename pc = PhysicalConstantsCGS>
+class ZhuTableOpacity {
+ public:
+  using PC = pc;
+  using DataBox = Spiner::DataBox<Real>;
+
+  ZhuTableOpacity() = default;
+
+  // construct Planck/Rosseland DataBox from Zhu ascii file
```
Yurlungur:
```suggestion
  // construct Planck/Rosseland DataBox from Zhu ascii or spiner hdf5 file
```

singularity-opac/photons/zhu_grey_table_opacity_photons.hpp
```diff
@@ -0,0 +1,309 @@
...
+    if (fexists) {
+
+      std::filesystem::path filePath(filename);
+      std::string extension = filePath.extension().string();
+
+      if (extension == ".txt") {
+        // assume this is one of the original Zhu et al (2021) ASCII files
+        loadZhuASCII(ff);
+#ifdef SPINER_USE_HDF
+      } else if (extension == ".hdf5" || extension == ".h5") {
```
Yurlungur:
```suggestion
      } else if (extension == ".hdf5" || extension == ".h5" || extension == ".sp5") {
```

singularity-opac/photons/zhu_grey_table_opacity_photons.hpp
```diff
@@ -0,0 +1,309 @@
...
+class ZhuTableOpacity {
+ public:
+  using PC = pc;
+  using DataBox = Spiner::DataBox<Real>;
+
+  ZhuTableOpacity() = default;
+
+  // construct Planck/Rosseland DataBox from Zhu ascii file
+  ZhuTableOpacity(const std::string filename, const bool use_planck_absorb = false)
+    : opac_type_(use_planck_absorb) {
```
Yurlungur: Do you plan to load both types at once? It might be possible to generate them both with only one file load. Maybe like this:

```C++
ZhuTableOpacity(const ZhuTableOpacity &other, const bool use_planck_absorb = false)
    : dist_(other.dist_), kappa_(other.kappa_), opac_type_(use_planck_absorb) {}
```
so you can initialize one variant from the other.

test/CMakeLists.txt
```diff
@@ -48,8 +48,17 @@ PRIVATE
...
+  test_zhu_grey_table_opacities.cpp
 )
 
+# Link data file directory for Zhu table test
+file(CREATE_LINK
+    ${CMAKE_SOURCE_DIR}/singularity-opac/photons/zhu_data/opacitysolar09dustq3p5amax0p1new.txt
+    ${CMAKE_CURRENT_BINARY_DIR}/opacitysolar09dustq3p5amax0p1new.txt SYMBOLIC)
+file(CREATE_LINK
+    ${CMAKE_SOURCE_DIR}/singularity-opac/photons/zhu_data/opacitysolar09dustq3p5amax0p1new.hdf5
+    ${CMAKE_CURRENT_BINARY_DIR}/opacitysolar09dustq3p5amax0p1new.hdf5 SYMBOLIC)
```
Yurlungur: If we use a data release model, this would be replaced with cmake fetch content.

Yurlungur (approved):
Looks good! Some minor suggestions but overall this seems like a pretty clean implementation. My main suggestion/complaint is to move the data files out of git. If we want to distribute them, we could maybe use github's release machinery. Let's discuss when you have the chance.

singularity-opac/photons/zhu_data/opacitysolar09dustq3p5amax0p1new.txt
RyanWollaeger: @Yurlungur - thanks very much for the comments - @brryan also recommended removing the files, and instead using a smaller "toy" test file - will make that change.

singularity-opac/photons/zhu_grey_table_opacity_photons.hpp
RyanWollaeger: Indeed, for use in transport specifically, was planning to permit even using both types at once in a hybrid-type scheme, which exists in the literature.  This constructor suggestion seems like a nice way to get both variants for instance, for grey diffusion. Will add it...

RyanWollaeger: Following @brryan's suggestion, this entire class has been removed, and its contents moved into `mean_opacity_photons.hpp`. There the Rosseland and Planck values naturally get accessed separately.

test/CMakeLists.txt
RyanWollaeger: Modified this to use a smaller example .txt file, but left the symlink as-is in the most recent change.

RyanWollaeger:
What do people think about slightly modifying the ASCII format: e.g. extending the header line with min/max density and temperature, so it can be parsed in rather than hard-coded?

Yurlungur:
> What do people think about slightly modifying the ASCII format: e.g. extending the header line with min/max density and temperature, so it can be parsed in rather than hard-coded?

I'm a bit nervous about doing that if the ASCII file is provided by someone else. However, if we assume control of it, that makes complete sense. You could also just demand the min/max densities be provided as inputs?

singularity-opac/photons/zhu_grey_table_opacity_photons.hpp
Yurlungur: ah that's a nice solution. :+1:

test/CMakeLists.txt
Yurlungur: the smaller file is certainly good enough IMO.

RyanWollaeger:
> > What do people think about slightly modifying the ASCII format: e.g. extending the header line with min/max density and temperature, so it can be parsed in rather than hard-coded?
> 
> I'm a bit nervous about doing that if the ASCII file is provided by someone else. However, if we assume control of it, that makes complete sense. You could also just demand the min/max densities be provided as inputs?

That's a fair point - right now, working on adding a preprocessing python script to generate a new file that: verifies the header data, adds the min/max rho-T values to the header calculated from the original file data, and optionally interpolates to a log-log grid in rho-T (if the grid is not log-log already).  This might help people vet their data, or provide a starting point.  However, we can defer modifying the ASCII that gets parsed in.

RyanWollaeger:
Just added the old->new ASCII preprocessing script in the last commit, but left the MeanOpacity class alone.

RyanWollaeger:
> Just added the old->new ASCII preprocessing script in the last commit, but left the MeanOpacity class alone.

Now the `loadASCII` function is modified to take in a modified format, which has min/max rho-T in the header and no longer has rho-T columns. The preprocessing script has SciPy's 2D interpolation function, so can remap any non-uniformly spaced data that has rho and T columns to a log-log format, as @Yurlungur originally suggested.

Yurlungur:
I'm happy with this. @RyanWollaeger @brryan merge when you feel its ready.

singularity-opac/photons/example_ascii/kap_plaw.txt
```diff
@@ -0,0 +1,19 @@
+ NRho 4 NT 4
+ rho_min 1.00000000e-14 rho_max 7.94328235e-01
```
brryan, line 2: Are these inclusive bounds, e.g. the first opacity value is at rho=1e-14 and the last is at rho=7.94...e-01, rather than the opacity values being "bin centers" where their `rho = exp(log(rho_min) + (i + 0.5)*dlogrho)` for the `i`th opacity value?

Would it be worth adding a short paragraph to the `README` explaining this format?

RyanWollaeger: These are indeed inclusive bounds. Will add a description to the README.

RyanWollaeger: Just added a section to the README.md file.

brryan: Looks great thanks!

singularity-opac/photons/example_ascii/preproc_ascii_opac.py
```diff
@@ -0,0 +1,71 @@
+#--------------------------------------------------------------------------------------------------#
```
brryan, line 16:
```suggestion
# ======================================================================
# © 2024. Triad National Security, LLC. All rights reserved.  This
# program was produced under U.S. Government contract
# 89233218CNA000001 for Los Alamos National Laboratory (LANL), which
# is operated by Triad National Security, LLC for the U.S.
# Department of Energy/National Nuclear Security Administration. All
# rights in the program are reserved by Triad National Security, LLC,
# and the U.S. Department of Energy/National Nuclear Security
# Administration. The Government is granted for itself and others
# acting on its behalf a nonexclusive, paid-up, irrevocable worldwide
# license in this material to reproduce, prepare derivative works,
# distribute copies to the public, perform publicly and display
# publicly, and to permit others to do so.
# ======================================================================

#--------------------------------------------------------------------------------------------------#
```

singularity-opac/photons/example_ascii/preproc_ascii_opac.py
```diff
@@ -0,0 +1,71 @@
...
+
+#-- save with new header containing min/max rho, T
+hdr = "NRho "+str(NRho)+" NT "+str(NT) + "\n" \
+    "rho_min {:.8e}".format(r[0])+" rho_max {:.8e}".format(r[NRho-1]) + "\n" \
+    "T_min {:.8e}".format(tt[0]) + " T_max {:.8e}".format(tt[NT-1])
+np.savetxt(args.fname_new, opac[:,2:], delimiter=" ", header=hdr, comments=" ", fmt="%.8e")
+
+#--------------------------------------------------------------------------------------------------#
+# End of preproc_ascii_opac.py
+#--------------------------------------------------------------------------------------------------#
```
brryan:
```suggestion
np.savetxt(args.fname_new, opac[:,2:], delimiter=" ", header=hdr, comments=" ", fmt="%.8e")

```

singularity-opac/photons/mean_opacity_photons.hpp
```diff
@@ -92,29 +113,27 @@ class MeanOpacity {
...
-    lkappaRosseland_.finalize();
+    lkappa_.finalize();
   }
 
   PORTABLE_INLINE_FUNCTION
   Real PlanckMeanAbsorptionCoefficient(const Real rho, const Real temp) const {
     Real lRho = toLog_(rho);
     Real lT = toLog_(temp);
-    return rho * fromLog_(lkappaPlanck_.interpToReal(lRho, lT));
+    return rho * fromLog_(lkappa_.interpToReal(lRho, lT, 1));
```
brryan: What do you think about adding a 
```
enum OpacityAveraging {
    Rosseland = 0,
    Planck = 1
};
```
so this can become `(lkappa_.interpToReal(lRho, lT, Planck)` here and elsewhere?

RyanWollaeger: It's a very good idea - should this `enum` be added here, or in the `base` subdirectory?

brryan (approved):
Few nitpicking comments, otherwise LGTM! Thanks for adding the test and documentation

singularity-opac/photons/mean_opacity_photons.hpp
brryan: I would add it here inside the `MeanOpacity` class to avoid polluting the namespace

RyanWollaeger: Just added the `enum` - thanks!

brryan: :+1:

PR #55, opened 2024-11-01
Missing PORTABLE_INLINE_FUNCTIONs that caused compiler warnings

brryan:
It was noticed that runtime physical constants were producing compiler warnings downstream.


Yurlungur (approved):
oops good catch

PR #56, opened 2024-11-20
Fix order of interpolation grids in preproc_ascii_opac.py (sorry).

RyanWollaeger:
## Background

The testing did not expose a bug in the preproc script, where the ordering of the rho-T grids was wrong in the interpolation function. This was due to testing on an input file that had an equal number of density and temperature points, along with equal opacity at each.

## Changes

* Fix rho, T grid ordering in SciPy `interp2d` function.
* Add an optional plotting capability, to look at color maps before and after interpolation:
  
Screen Shot 2024-11-20 at 12 43 28 PM


RyanWollaeger:
@brryan @Yurlungur - sorry to have not checked this script interpolation more carefully initially.

Yurlungur (approved):
Nah easy fix. Thanks for getting on it so quickly.

PR #57, opened 2024-11-20
nulib reader

Yurlungur:
Reads nulib data. Needs my fork of nulib, see: https://github.com/Yurlungur/NuLib/pull/1


AstroBarker (approved):
Thanks for this! It looks good to me, I think. I hack-tested it by putting a call to `LoadFromNuLib_` with a nulib table in the test constructor and it didn't crash.

Yurlungur:
Good enough for me. Thanks @AstroBarker . Gonna merge this. Then at some point one of us should put it into phoebus for real.

PR #58, opened 2024-12-10
Non-CGS constants

brryan:
We want to be able to extract runtime physical constants in the unit system of a non-cgs opacity when using non-cgs opacities. 

- [x] Remove some probably irrelevant constants
- [x] Update photons
- [x] Update neutrinos


brryan:
OK this is ready for review, and merge if it looks good. Note that consistency between singularity-opac unit system and units used in data tables read in is still the responsibility of the downstream user.

Yurlungur (approved):
LGTM

pdmullen (approved):
Sorry for the delayed review.  All LGTM!

PR #59, opened 2024-12-17
Add AbsorptionCoefficient and Emissivity to MeanOpacity variants.

RyanWollaeger:
+ Use gmode input to toggle Rosseland (default) to Planck.
+ Add PlanckDistribution function as member to MeanOpacity.
+ Add the new functions to generic and non-CGS MeanOpacity variants.

Note: the motivation is to use one of the frequency-integrated opacity values as a grey opacity in transport, which requires using it in both absorption and emissivity, for any choice.

### DIFF
--- a/singularity-opac/photons/mean_opacity_photons.hpp
+++ b/singularity-opac/photons/mean_opacity_photons.hpp
@@ -29,6 +29,7 @@
 
 #include <singularity-opac/photons/mean_photon_variant.hpp>
 #include <singularity-opac/photons/non_cgs_photons.hpp>
+#include <singularity-opac/photons/thermal_distributions_photons.hpp>
 
 namespace singularity {
 namespace photons {
@@ -141,6 +142,25 @@ class MeanOpacity {
     Real lRho = toLog_(rho);
     Real lT = toLog_(temp);
     return rho * fromLog_(lkappa_.interpToReal(lRho, lT, Rosseland));
+  }
+
+  // standard grey opacity functions
+  PORTABLE_INLINE_FUNCTION
+  Real AbsorptionCoefficient(const Real rho, const Real temp,
+                            const int gmode = Rosseland) const {
+    Real lRho = toLog_(rho);
+    Real lT = toLog_(temp);
+    return rho * fromLog_(lkappa_.interpToReal(lRho, lT, gmode));
+  }
+
+  PORTABLE_INLINE_FUNCTION
+  Real Emissivity(const Real rho, const Real temp,
+                  const int gmode = Rosseland,
+                  Real *lambda = nullptr) const {
+    Real B = dist_.ThermalDistributionOfT(temp, lambda);
+    Real lRho = toLog_(rho);
+    Real lT = toLog_(temp);
+    return rho * fromLog_(lkappa_.interpToReal(lRho, lT, gmode)) * B;
   }
 
  private:
@@ -307,6 +327,7 @@ class MeanOpacity {
     Rosseland = 0,
     Planck = 1
   };
+  PlanckDistribution<PC> dist_;
 };
 
 #undef EPS
--- a/singularity-opac/photons/mean_opacity_photons.hpp
+++ b/singularity-opac/photons/mean_opacity_photons.hpp
@@ -323,10 +323,6 @@ class MeanOpacity {
   }
   Spiner::DataBox<Real> lkappa_;
   const char *filename_;
-  enum OpacityAveraging {
-    Rosseland = 0,
-    Planck = 1
-  };
   PlanckDistribution<PC> dist_;
 };
 
--- a/singularity-opac/photons/mean_photon_variant.hpp
+++ b/singularity-opac/photons/mean_photon_variant.hpp
@@ -21,18 +21,13 @@
 #include <ports-of-call/portability.hpp>
 #include <singularity-opac/base/opac_error.hpp>
 #include <singularity-opac/base/radiation_types.hpp>
+#include <singularity-opac/photons/mean_photon_types.hpp>
 #include <singularity-opac/photons/photon_variant.hpp>
 #include <variant/include/mpark/variant.hpp>
 
 namespace singularity {
 namespace photons {
 namespace impl {
-
-// mean-opacity mode
-enum OpacityAveraging {
-  Rosseland = 0,
-  Planck = 1
-};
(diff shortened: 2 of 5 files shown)


Yurlungur (approved):
This may eventually be something we want for neutrinos too... in which case, it might be good to add it there too. The API should be identical. That said, if you don't want to attack that right now, that's fine.

RyanWollaeger:
@brryan @Yurlungur @pdmullen - this is a preliminary attempt for something that might be used in grey transport. I added the Planck function as a member to the MeanOpacity class, which seems a bit overkill for just emissivity.

singularity-opac/photons/mean_opacity_photons.hpp
```diff
@@ -143,6 +144,25 @@ class MeanOpacity {
     return rho * fromLog_(lkappa_.interpToReal(lRho, lT, Rosseland));
   }
 
+  // standard grey opacity functions
+  PORTABLE_INLINE_FUNCTION
+  Real AbsorptionCoefficient(const Real rho, const Real temp,
+                            const int gmode = Rosseland) const {
```
RyanWollaeger, line 150: Would it make sense to move this `gmode` index to the constructor? We may want to avoid different function signatures between non-grey mean opacities and mean opacities, so that they can be used interchangeably.

singularity-opac/photons/mean_photon_variant.hpp
```diff
@@ -92,6 +92,27 @@ class MeanVariant {
         opac_);
   }
 
+  PORTABLE_INLINE_FUNCTION
+  Real AbsorptionCoefficient(const Real rho, const Real temp,
+                            const int gmode = 0) const {
```
brryan:
```suggestion
                            const int gmode = Rosseland) const {
```
Is this change OK to make? There are a few other places with integers as default args to these functions here and below.

brryan (approved):
LGTM! I think the planck/rosseland choice is best made when calling the function, either as you do here with function args, or with different fucntions or a template parameter. What you have here is fine IMO, I just think it's better to not do this in the constructor because we dont want to make the same class twice to call the same data. I'm unconcerned about the different function signatures between mean and non-mean opacity variants, because they have different meanings.

singularity-opac/photons/mean_photon_variant.hpp
RyanWollaeger: For some reason my comment got duplicated, I deleted the duplication, but that deleted it here - I made the change, but asked about `non_cgs_photons.hpp`, which doesn't include this as a header, but has these `gmode` values also defaulted to integers. Should the enum be moved to a separate header to be included here and in `non_cgs_photons.hpp`?

Yurlungur:
@RyanWollaeger what's the status of this MR?

singularity-opac/photons/mean_opacity_photons.hpp
RyanWollaeger: We decided this should be in the AbsorptionCoefficient and Emissivity arguments.

RyanWollaeger: (Accidentally added this comment in a pending review)

singularity-opac/photons/mean_photon_variant.hpp
RyanWollaeger: @brryan - any thoughts on this header question? Thanks!

brryan: @RyanWollaeger Sorry I'm not sure how I missed this question. Yeah I think you're right, we need a new header file, `photon_types.hpp` or something (you may have a better name in mind) to be included in both these existing files.

singularity-opac/photons/non_cgs_photons.hpp
```diff
@@ -246,7 +246,8 @@ class MeanNonCGSUnits {
                   const Real temp_unit)
       : mean_opac_(std::forward<MeanOpac>(mean_opac)), time_unit_(time_unit),
         mass_unit_(mass_unit), length_unit_(length_unit), temp_unit_(temp_unit),
-        rho_unit_(mass_unit_ / (length_unit_ * length_unit_ * length_unit_)) {}
+        rho_unit_(mass_unit_ / (length_unit_ * length_unit_ * length_unit_)),
+        inv_emiss_unit_(length_unit_ * time_unit_ * time_unit_ / mass_unit_) {}
```
brryan: Is this not 
```suggestion
        inv_emiss_unit_(length_unit_ * time_unit_ * time_unit_ * time_unit_ / mass_unit_) {}
```
? since
`<J> = <dE / d^3x dt> = g cm^2 / (s^2 cm^3 s) = g / (cm s^3)`

brryan: Oh I see you add the extra `time_unit_` in separately later... but any reason to not just include that here?

singularity-opac/photons/mean_photon_variant.hpp
Yurlungur: Should this be mimicked in neutrinos too? It might be useful to have rosseland vs planck mean in neutrinos too

singularity-opac/photons/non_cgs_photons.hpp
RyanWollaeger: This may've been to conform to the pattern for the non-mean variant - will double-check.

singularity-opac/photons/mean_photon_variant.hpp
RyanWollaeger: Just made this update for photons - will map this over to neutrinos shortly, along with the other AbsorptionCoefficient/Emissivity changes.

singularity-opac/photons/non_cgs_photons.hpp
RyanWollaeger: Just made this change - thanks!

brryan: 🙌

singularity-opac/photons/mean_photon_variant.hpp
brryan: We can try out just the photons for now and see if anything goes wrong bringing it into artemis before typing out the neutrinos

brryan (approved):
LGTM! Great work @RyanWollaeger!

PR #60, opened 2025-01-02
Fix mean opacities

Yurlungur:
In the recent MR #58 where runtime physical constants were added, the mean opacities needed to be changed to not look for the compile time constants. I make that modification here. @AstroBarker saw this when trying to update singularity-opac in phoebus.

I know we are likely going to make more changes here soon to support Artemis + Riot but I think it's worth fixing this now, and then we can iterate later as needed.

### DIFF
--- a/singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
+++ b/singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
@@ -140,9 +140,11 @@ class MeanOpacity {
                         const Real lTMax, const int NT, const Real YeMin,
                         const Real YeMax, const int NYe, Real lNuMin,
                         Real lNuMax, const int NNu, Real *lambda = nullptr) {
-    static_assert(std::is_same<typename Opacity::PC, PC>::value,
-                  "Error: MeanOpacity physical constants do not match those of "
-                  "Opacity used for construction.");
+#ifndef NDEBUG
+    auto RPC = RuntimePhysicalConstants(PC);
+    auto opc = opac.GetRuntimePhysicalConstants();
+    assert(RPC == opc && "Physical constants are the same");
+#endif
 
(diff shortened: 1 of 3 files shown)


AstroBarker:
Thanks for going after this! It doesn't seem to be compiling for me with a Debug build with several `expected primary expression before (` pointing to, e.g., `mean_opacity_neutrinos.hpp:144`, `auto RPC = RuntimePhysicalConstants(PC)`. Also, the mean neutrino opacity test fails locally for me.

singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
```diff
@@ -140,9 +140,11 @@ class MeanOpacity {
                         const Real lTMax, const int NT, const Real YeMin,
                         const Real YeMax, const int NYe, Real lNuMin,
                         Real lNuMax, const int NNu, Real *lambda = nullptr) {
-    static_assert(std::is_same<typename Opacity::PC, PC>::value,
-                  "Error: MeanOpacity physical constants do not match those of "
-                  "Opacity used for construction.");
+#ifndef NDEBUG
+    auto RPC = RuntimePhysicalConstants(PC);
```
AstroBarker: Should this be `PC()`? And similar below?

Yurlungur:
> Thanks for going after this! It doesn't seem to be compiling for me with a Debug build with several `expected primary expression before (` pointing to, e.g., `mean_opacity_neutrinos.hpp:144`, `auto RPC = RuntimePhysicalConstants(PC)`. Also, the mean neutrino opacity test fails locally for me.

Oops my bad... Didn't run tests. Need to call the constructor in that check. Try now.

AstroBarker (approved):
Compiling and all tests passing locally now. Thanks!

brryan (approved):
LGTM thanks for catching this

PR #63, opened 2025-01-16
Modify evaluation of absorption coefficients

pdmullen:
This PR originally started much larger---actually changing all calls from `AbsorptionCoefficient` to `SpecificAbsorptionOpacity`.  After some retrospection, a simple change of mindset got me on board with instead dealing with absorption coefficients.  I'd like to discuss this via meeting.

After this realization, this PR shrunk to only changing out some calls to `AbsorptionCoefficient` to directly returning rho * kappa (where rho is the material density and kappa is the material specific absorption opacity), instead of arriving there via Kirchhoff's laws.  

I kept the functions for `*FromKirchhoff` in this PR for photons, even though they are not used, as I think they do help guide readers of the code as to what things correspond to.  Neutrinos also have calls to `*FromKirchhoff` (e.g., for the tophat absorption coefficient), but I did not edit these (was this the right move?).

### DIFF
--- a/singularity-opac/neutrinos/tophat_emissivity_neutrinos.hpp
+++ b/singularity-opac/neutrinos/tophat_emissivity_neutrinos.hpp
@@ -55,7 +55,7 @@ class TophatEmissivity {
   Real AbsorptionCoefficient(const Real rho, const Real temp, const Real Ye,
                              const RadiationType type, const Real nu,
                              Real *lambda = nullptr) const {
-    return dist_.AbsorptionCoefficientFromKirkhoff(*this, rho, temp, Ye, type,
+    return dist_.AbsorptionCoefficientFromKirchhoff(*this, rho, temp, Ye, type,
                                                    nu, lambda) /
            (4. * M_PI);
   }
@@ -78,7 +78,7 @@ class TophatEmissivity {
                                           const RadiationType type,
                                           const Real nu,
                                           Real *lambda = nullptr) const {
-    return dist_.AngleAveragedAbsorptionCoefficientFromKirkhoff(
+    return dist_.AngleAveragedAbsorptionCoefficientFromKirchhoff(
         *this, rho, temp, Ye, type, nu, lambda);
   }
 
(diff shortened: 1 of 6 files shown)


singularity-opac/neutrinos/tophat_emissivity_neutrinos.hpp
```diff
@@ -55,7 +55,7 @@ class TophatEmissivity {
   Real AbsorptionCoefficient(const Real rho, const Real temp, const Real Ye,
                              const RadiationType type, const Real nu,
                              Real *lambda = nullptr) const {
-    return dist_.AbsorptionCoefficientFromKirkhoff(*this, rho, temp, Ye, type,
+    return dist_.AbsorptionCoefficientFromKirchhoff(*this, rho, temp, Ye, type,
```
Yurlungur, line 58: oops thank you for the spelling fix...

singularity-opac/photons/gray_opacity_photons.hpp
```diff
@@ -49,8 +49,7 @@ class GrayOpacity {
   PORTABLE_INLINE_FUNCTION
   Real AbsorptionCoefficient(const Real rho, const Real temp, const Real nu,
                              Real *lambda = nullptr) const {
-    return dist_.AbsorptionCoefficientFromKirkhoff(*this, rho, temp, nu,
-                                                   lambda);
+    return rho * kappa_;
```
Yurlungur, line 52: :+1:

Yurlungur (approved):
LGTM. Thanks for the cleanup! Yes let's discuss the larger picture changes together at some later date.

Yurlungur:
> I kept the functions for *FromKirchhoff in this PR for photons, even though they are not used, as I think they do help guide readers of the code as to what things correspond to. 

I agree with this choice

> Neutrinos also have calls to *FromKirchhoff (e.g., for the tophat absorption coefficient), but I did not edit these (was this the right move?).

I think it's maybe worth editing those down the line but no need to change it here. :+1:

PR #66, opened 2025-05-01
Make slight tweaks for Debug builds.

RyanWollaeger:
### Description

Testing with a Debug build of a dependent code revealed a comparison of the same constants throws an assertion.

### Changes

+ Modify memcmp usage in == overload comparing physical constants.
+ Update spiner submodule for fix of a recently added assert.

### DIFF
--- a/singularity-opac/constants/constants.hpp
+++ b/singularity-opac/constants/constants.hpp
@@ -236,7 +236,7 @@ struct RuntimePhysicalConstants {
 
   // Runtime physical constants are trivially copyable
   bool operator==(const RuntimePhysicalConstants &other) const {
-    return std::memcmp(this, &other, sizeof(*this));
+    return (std::memcmp(this, &other, sizeof(*this)) == 0);
   }
 };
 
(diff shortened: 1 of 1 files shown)


Yurlungur (approved):

RyanWollaeger:
@Yurlungur would it be reasonable to convert the bit-wise comparison in the overload to comparing the member data with tolerances (to permit, e.g. 1e-15 differences in units)? Or is that maybe not in the spirit of this comparison?

Yurlungur:
> @Yurlungur would it be reasonable to convert the bit-wise comparison in the overload to comparing the member data with tolerances (to permit, e.g. 1e-15 differences in units)? Or is that maybe not in the spirit of this comparison?

I think that would be fine. The current implementation is just laziness on my part I think.

RyanWollaeger:
> > @Yurlungur would it be reasonable to convert the bit-wise comparison in the overload to comparing the member data with tolerances (to permit, e.g. 1e-15 differences in units)? Or is that maybe not in the spirit of this comparison?
> 
> I think that would be fine. The current implementation is just laziness on my part I think.

Will make that change quickly here then - thanks!

RyanWollaeger:
Just made the eps-comparison modification (the memcmp is much more succinct).

Yurlungur (approved):

PR #67, opened 2025-05-09
Another modification to constants comparison

adamdempsey90:
Debug build of artemis crashed in the constants comparison (of `h` specifically). Seemed like a stupid floating point problem that was solved by making the comparison `<=` instead of `<`. I also went ahead and split out the difference and scale into separate variables for possible future debugging.

### DIFF
--- a/singularity-opac/constants/constants.hpp
+++ b/singularity-opac/constants/constants.hpp
@@ -265,10 +265,12 @@ struct RuntimePhysicalConstants {
     return is_same;
   }
 
-  private:
-   bool eps_equal(const Real &a, const Real &b) const {
-     return 2.0 * std::abs(a - b) < EPS * (std::abs(a) + std::abs(b) + 1e-100);
-   }
+ private:
+  bool eps_equal(const Real &a, const Real &b) const {
+    const Real diff = std::abs(a - b);
+    const Real scale = 0.5 * (std::abs(a) + std::abs(b));
+    return diff / (scale + 1e-100) <= EPS;
+  }
 };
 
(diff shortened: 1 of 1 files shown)


singularity-opac/constants/constants.hpp
```diff
@@ -265,10 +265,12 @@ struct RuntimePhysicalConstants {
...
 
-  private:
-   bool eps_equal(const Real &a, const Real &b) const {
-     return 2.0 * std::abs(a - b) < EPS * (std::abs(a) + std::abs(b) + 1e-100);
-   }
+ private:
+  bool eps_equal(const Real &a, const Real &b) const {
+    const Real diff = std::abs(a - b);
+    const Real scale = 0.5 * (std::abs(a) + std::abs(b));
+    return diff / (scale + 1e-100) <= EPS;
```
Yurlungur, line 272: I'm a little surprised this is more robust.

Yurlungur (approved):

singularity-opac/constants/constants.hpp
adamdempsey90: Maybe something to do with storing the result of the operations before the comparison to eps? 
https://godbolt.org/z/vc848vMds

adamdempsey90: Or doing  0 < EPS * 1e-27 is not well defined (in the sense the RHS is set to 0?)

Yurlungur: yeah I think not dividing is maybe technically better. But changing < to <= and storing the pieces could for sure matter.

PR #69, opened 2026-04-09
Fix thermal distribution bugs

pdmullen:
This PR used generative AI.

Fixes the photon and neutrino dB_\nu/dT implementations used for Rosseland weighting.  Adds focused thermal-distribution regression coverage and verifies the fix against test suites.


test/CMakeLists.txt
```diff
@@ -10,6 +10,8 @@
 # prepare derivative works, distribute copies to the public, perform
 # publicly and display publicly, and to permit others to do so.
 
+# This file was made in part with generative AI.
+
```
Yurlungur, line 14: probably not needed in this file (yet) lol

test/test_mean_opacities.cpp
```diff
@@ -72,11 +74,11 @@ TEST_CASE("Mean neutrino opacities", "[MeanNeutrinos]") {
     constexpr Real nu = 1.25 * MeV2Hz; // 1 MeV
 
     constexpr int nT = 10;
-    constexpr Real lRhoMin = std::log10(0.1 * rho);
-    constexpr Real lRhoMax = std::log10(10. * rho);
+    const Real lRhoMin = std::log10(0.1 * rho);
+    const Real lRhoMax = std::log10(10. * rho);
```
Yurlungur, line 78: Does this need to be changed? It looks like it might have gotten caught up in a larger automated formatting sweep.

test/test_spiner_opac_neutrinos.cpp
```diff
@@ -60,8 +60,8 @@ TEST_CASE("Spiner opacities, filled with gray data",
   constexpr Real lRhoMin = 8;
   constexpr Real lRhoMax = 12;
   constexpr int NRho = 32;
-  constexpr Real lTMin = -2 + std::log10(MeV2K);
-  constexpr Real lTMax = 2 + std::log10(MeV2K);
+  const Real lTMin = -2 + std::log10(MeV2K);
+  const Real lTMax = 2 + std::log10(MeV2K);
```
Yurlungur, line 64: I think this should be reverted to constexpr.

Yurlungur (approved):
omg I can't believe this error was sitting in the code for so long. Thanks for the fix.

test/CMakeLists.txt
pdmullen: I know it sounds extremely lazy, but I also had generative AI fix my dates for me 🫣.  So maybe this is still required?

test/test_mean_opacities.cpp
pdmullen: I think it does.  I was actually seeing compile errors.  Looks like `std::log10` `constexpr` is not supported until C++26.  Crazy!

https://en.cppreference.com/w/cpp/numeric/math/log10

test/CMakeLists.txt
Yurlungur: 🤷 its probably fine either way. Fine to leave it.

test/test_mean_opacities.cpp
Yurlungur: huh. Wild it worked until now. Thanks for the catch then! My bad.

test/test_spiner_opac_neutrinos.cpp
Yurlungur: Incorrect on my part. My bad.

test/CMakeLists.txt
lroberts36: We would probably save ourselves some time if we just wrote a script that puts it at the top of every file...

Yurlungur: you're not wrong.

RyanWollaeger (approved):

PR #70, opened 2026-04-09
Add optional frequency dependence to PowerLaw

pdmullen:
This PR used generative AI.

Adds optional frequency dependence to the `PowerLaw` opacity.

### DIFF
--- a/singularity-opac/photons/powerlaw_opacity_photons.hpp
+++ b/singularity-opac/photons/powerlaw_opacity_photons.hpp
@@ -36,16 +36,39 @@ class PowerLawOpacity {
 
   PowerLawOpacity() = default;
   PowerLawOpacity(const Real kappa0, const Real rho_exp, const Real temp_exp,
-                  const Real nu_exp = 0., const Real nu_ref = 1.)
+                  const Real rho_ref = 1., const Real rho_off = 0.,
+                  const Real temp_ref = 1., const Real temp_off = 0.,
+                  const Real nu_exp = 0., const Real nu_ref = 1.,
+                  const Real nu_off = 0., const bool do_stim_emit = false)
       : PowerLawOpacity(PlanckDistribution<pc>{}, kappa0, rho_exp, temp_exp,
-                        nu_exp, nu_ref) {}
+                        rho_ref, rho_off, temp_ref, temp_off,
+                        nu_exp, nu_ref, nu_off, do_stim_emit) {}
   PowerLawOpacity(const PlanckDistribution<pc> &dist, const Real kappa0,
-                  const Real rho_exp, const Real temp_exp, const Real nu_exp = 0.0,
-                  const Real nu_ref = 1.)
+                  const Real rho_exp, const Real temp_exp,
+                  const Real rho_ref = 1., const Real rho_off = 0.,
+                  const Real temp_ref = 1., const Real temp_off = 0.,
+                  const Real nu_exp = 0., const Real nu_ref = 1., const Real nu_off = 0.,
+                  const bool do_stim_emit = false)
       : dist_(dist), kappa0_(kappa0), rho_exp_(rho_exp), temp_exp_(temp_exp),
-        nu_exp_(nu_exp), nu_ref_(nu_ref) {
+        rho_ref_(rho_ref), rho_off_(rho_off), temp_ref_(temp_ref), temp_off_(temp_off),
+        nu_exp_(nu_exp), nu_ref_(nu_ref), nu_off_(nu_off), do_stim_emit_(do_stim_emit) {
+    if (!(rho_ref_ > 0.)) {
+      OPAC_ERROR("PowerLawOpacity: rho_ref must be positive");
+    }
+    if (!(temp_ref_ > 0.)) {
+      OPAC_ERROR("PowerLawOpacity: temp_ref must be positive");
+    }
     if (!(nu_ref_ > 0.)) {
       OPAC_ERROR("PowerLawOpacity: nu_ref must be positive");
+    }
+    if (rho_off_ < 0.) {
+      OPAC_ERROR("PowerLawOpacity: rho_off must be nonnegative");
+    }
+    if (temp_off_ < 0.) {
+      OPAC_ERROR("PowerLawOpacity: temp_off must be nonnegative");
+    }
+    if (nu_off_ < 0.) {
+      OPAC_ERROR("PowerLawOpacity: nu_off must be nonnegative");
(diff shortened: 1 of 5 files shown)


Yurlungur (approved):
looks good

singularity-opac/photons/powerlaw_opacity_photons.hpp
```diff
@@ -178,9 +206,21 @@ class PowerLawOpacity {
   }
 
  private:
-  Real kappa0_;   // Opacity scale. Units of cm^2/g
+  PORTABLE_INLINE_FUNCTION
+  Real OpacityScale_(const Real rho, const Real temp, const Real nu) const {
+    return OpacityPrefactor_(rho, temp) * std::pow(nu / nu_ref_, nu_exp_);
```
RyanWollaeger: Would it make sense to add some sort of regularization parameter, so that the opacity plateaus at some particular frequency rather than continue up or down?

singularity-opac/photons/powerlaw_opacity_photons.hpp
```diff
@@ -33,26 +35,35 @@ class PowerLawOpacity {
...
-  PowerLawOpacity(const Real kappa0, const Real rho_exp, const Real temp_exp)
-      : kappa0_(kappa0), rho_exp_(rho_exp), temp_exp_(temp_exp) {}
+  PowerLawOpacity(const Real kappa0, const Real rho_exp, const Real temp_exp,
+                  const Real nu_exp = 0., const Real nu_ref = 1.)
+      : PowerLawOpacity(PlanckDistribution<pc>{}, kappa0, rho_exp, temp_exp,
+                        nu_exp, nu_ref) {}
   PowerLawOpacity(const PlanckDistribution<pc> &dist, const Real kappa0,
-                  const Real rho_exp, const Real temp_exp)
-      : dist_(dist), kappa0_(kappa0), rho_exp_(rho_exp), temp_exp_(temp_exp) {}
+                  const Real rho_exp, const Real temp_exp, const Real nu_exp,
```
RyanWollaeger: Should this have `nu_exp = 0` as well? (Maybe missing something here)

RyanWollaeger:
LGTM - it'd be nice to have this in soon, for testing in dependent code.

RyanWollaeger:
Sorry to jump the gun on the dBdT fix - it looks like this PR has to be redirected to `main` now

singularity-opac/photons/powerlaw_opacity_photons.hpp
RyanWollaeger: Just added positive offsets (not really the same thing, but similar purpose) for frequency, density and temperature.

test/test_powerlaw_opacities.cpp
```diff
@@ -196,3 +200,119 @@ TEST_CASE("CGS power law photon opacities", "[PowerLawCGSPhotonOpacities]") {
...
+  constexpr Real rho = 1.5e0;    // g/cc
+  constexpr Real temp = 1.e3;    // K
+  constexpr Real nu_min = 1.e10; // Hz
+  constexpr Real nu_max = 1.e14; // Hz
+  constexpr int n_nu = 100;
+  constexpr Real kappa0 = 1.5;   // cm^2 / g / Hz^nu_exp
+  constexpr Real nu_ref = 1.e12; // Hz
+
+  WHEN("We initialize a frequency-dependent CGS power law photon opacity") {
+    photons::PowerLaw opac_host(kappa0, rho_exp, temp_exp, nu_exp, nu_ref);
```
RyanWollaeger, line 215: Sorry - will update these tests or switch the ctor argument ordering shortly.

singularity-opac/photons/powerlaw_opacity_photons.hpp
```diff
@@ -208,19 +231,29 @@ class PowerLawOpacity {
  private:
   PORTABLE_INLINE_FUNCTION
   Real OpacityScale_(const Real rho, const Real temp, const Real nu) const {
-    return OpacityPrefactor_(rho, temp) * std::pow(nu / nu_ref_, nu_exp_);
+    const Real freq_plaw = std::pow((nu + nu_off_) / nu_ref_, nu_exp_);
+    const Real stim_fact = do_stim_emit_ ? -std::expm1(-(pc::h * nu / (pc::kb * temp))) : 1.0;
```
pdmullen, line 238: this brings in units... can you double check that all the conversions are threaded through correctly to get the correctly unit-informed Planck and Boltzmann constant?

RyanWollaeger: Good catch - will do...

RyanWollaeger: Since this is the CGS side of the class, it seems `nu` and `temp` should be in CGS by the time they reach this function, so that leaves the question of whether `pc::h` and `pc::kb` are properly non-dimensionalizing the exponential argument (or equivalently, what the actual units of `nu` is - e.g. energy or 1/time). If the `nu` here is assumed consistent with the `nu` taken by the `EmissivityPerNuPerOmega` function, it seems this should be correct: `EmissivityPerNuPerOmega` calls `ThermalDistributionOfNu` which takes `nu` directly and computes the nondimensional parameter `x = pc::h * nu / (pc::kb * temp)` as in the stimulated emission formula here. Let me know if this reasoning misses something.

Yurlungur:
@pdmullen @RyanWollaeger is this good to go? It looks like it is... and apparently I already approved ;)

RyanWollaeger:
> @pdmullen @RyanWollaeger is this good to go? It looks like it is... and apparently I already approved ;)

We're making a few final changes now (adding a test for the scattering power law opacity, minor comment updates)

RyanWollaeger:
It should be ready now, if things look okay - the new test is not exhaustive for the constructor arguments, but maybe that is not too concerning.

singularity-opac/photons/powerlaw_opacity_photons.hpp
RyanWollaeger: Following on this, noticed that the parameters `rho_ref` and `temp_ref` must be in CGS when the constructor is invoked. That is maybe a bit cumbersome - noted this in the comments next to these class members in each of the power law classes (also made a note that `nu_ref` has to have units of 1/s, which I think is true for the same reason unless `kappa0` absorbs the unit).

Yurlungur (approved):
I think it's fine. Thanks @RyanWollaeger !

PR #71, opened 2026-04-09
Towards Multigroup API

pdmullen:
_This PR and PR summary text used generative AI._

### Add photon multigroup absorption & scattering API

  Adds a multigroup radiation API for photons by unifying multigroup support into the existing mean-opacity classes
  (absorption and scattering) rather than adding new types. With ngroups=1 and bounds [0, ∞) these reduce exactly to
  gray opacities; ngroups>1 enables multigroup transport.

  - Group coefficients on (rho, T, group), built from a monochromatic model + group bounds or loaded from Spiner/SP5
  tables
  - Half-open group bounds [nu_g, nu_{g+1}) with lower/upper tail groups (0. and IEEE +inf); frequency-entry APIs
  (GroupOfNu, ...FromNu); non-CGS wrappers preserved
  - Shared helpers factored into mean_photon_utils.hpp; removed leftover ASCII loading

  Full suite passes with tests + HDF5 on (686 assertions, 36 cases). Deferred to follow-ups: neutrino multigroup,
  integrated multigroup emissivity.


Yurlungur:
@pdmullen is this ready for review?

plan_histories/71.md
```diff
@@ -0,0 +1,36 @@
...
+  - autogenerated logarithmic groups with tail groups
+  - direct table-backed construction
+  - SP5/HDF round-trip
+  - half-open frequency-bound behavior
+  - non-CGS wrapper behavior
+
+## Scope Boundaries
+
+- Photon multigroup remains absorption-only for now (scattering in later PR).
+- Do not add neutrino multigroup support in this PR (coming in later PR).
```
RyanWollaeger, line 36: At a glance it looks like these two scope boundaries are no longer true?

singularity-opac/photons/mean_opacity_photons.hpp
```diff
@@ -44,286 +49,480 @@ template <typename pc = PhysicalConstantsCGS>
...
+  MeanOpacity(const DataBox &kappaPlanck, const DataBox &kappaRosseland,
+                    const GroupBoundsIndexer &group_bounds) {
+    // Table-backed multigroup opacities always carry explicit group bounds.
+    // To represent [nu_max, infinity), the final bound must literally be
+    // IEEE +infinity, not a large finite proxy value.
+    LoadOpacityTables_(kappaPlanck, kappaRosseland, group_bounds);
+  }
 
-      if (extension == ".txt") {
-        loadASCII(ff);
```
RyanWollaeger, line 78: It looks like this breaks ASCII table parsing - is this something we should no longer support? (That seems reasonable - it is just going to affect downstream codes.)

test/test_mean_opacities.cpp
```diff
@@ -1,872 +0,0 @@
...
-            Real alphaRosselandFunny =
-                funny_units.RosselandMeanTotalScatteringCoefficient(
-                    rho / rho_unit, temp / temp_unit);
-            if (FractionalDifference(alphaPlanck, alphaPlanckFunny /
-                                                      length_unit) > EPS_TEST) {
-              n_wrong_d() += 1;
-            }
-            if (FractionalDifference(alphaRosseland,
-                                     alphaRosselandFunny / length_unit) >
-                EPS_TEST) {
```
RyanWollaeger, line 772: This `kap_plaw.txt` example file can probably be deleted now - think it is just used for this test.

singularity-opac/photons/mean_opacity_photons.hpp
```diff
@@ -44,286 +49,480 @@ template <typename pc = PhysicalConstantsCGS>
...
+  Real fromLog_(const Real lx) const { return std::pow(10., lx); }
+
+  PORTABLE_INLINE_FUNCTION
+  int GroupOfNuImpl_(const Real nu) const {
+    if (nu == GroupBoundAt_(groupBounds_, ngroups_)) {
+      return ngroups_ - 1;
+    }
+    int lower = 0;
+    int upper = ngroups_;
+    while (upper - lower > 1) {
```
RyanWollaeger: Might be worth noting in a comment that this is a binary search

singularity-opac/photons/mean_opacity_photons.hpp
```diff
@@ -44,286 +49,480 @@ template <typename pc = PhysicalConstantsCGS>
...
+  Real toLog_(const Real x) const { return std::log10(std::abs(x) + EPS); }
+
+  PORTABLE_INLINE_FUNCTION
+  Real fromLog_(const Real lx) const { return std::pow(10., lx); }
+
+  PORTABLE_INLINE_FUNCTION
+  int GroupOfNuImpl_(const Real nu) const {
+    if (nu == GroupBoundAt_(groupBounds_, ngroups_)) {
+      return ngroups_ - 1;
+    }
```
RyanWollaeger: Is there a similar return shortcut for `nu` at the lowest group bound (also should `nu` be compared by `==` to the group bound, or might this be more robust as an inequality)?

RyanWollaeger:
Looks like `test_mean_opacities.cpp` got modified on `main` - it should be straightforward to resolve that branch conflict.

RyanWollaeger:
Since the gray MeanOpacity class has been extended to multigroup here, does it make sense to support overloads of the opacity accessors that don't take a group index in the event that there is only one group (i.e. the tables are gray)? Or should that be a concern of the downstream codes?

singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
```diff
@@ -174,33 +174,32 @@ class MeanOpacity {
...
             }
-            const Real dlnu = (lNuMax - lNuMin) / (NNu - 1);
-            // Integrate over frequency
+            const Real dlnu = (lNuMax - lNuMin) / NNu;
+            // Integrate over frequency using midpoint rule
             for (int inu = 0; inu < NNu; ++inu) {
-              const Real weight =
-                  (inu == 0 || inu == NNu - 1) ? 0.5 : 1.; // Trapezoidal rule
-              const Real lnu = lNuMin + inu * dlnu;
+              const Real lnu = lNuMin + (inu + 0.5) * dlnu;
```
Yurlungur, line 180: no objections to this change here, but I'm curious about the change. Why the shift? Was the original wrong?

singularity-opac/photons/mean_opacity_photons.hpp
Yurlungur: I'm gonna defer to you guys on this one.

singularity-opac/photons/mean_opacity_photons.hpp
```diff
@@ -44,286 +49,480 @@ template <typename pc = PhysicalConstantsCGS>
...
+
+  PORTABLE_INLINE_FUNCTION
+  bool HasGroupBounds() const noexcept { return true; }
+
   PORTABLE_INLINE_FUNCTION
   Real PlanckMeanAbsorptionCoefficient(const Real rho, const Real temp) const {
-    Real lRho = toLog_(rho);
-    Real lT = toLog_(temp);
-    return rho * fromLog_(lkappa_.interpToReal(lRho, lT, Planck));
+    return PlanckGroupAbsorptionCoefficient(rho, temp, 0);
```
Yurlungur, line 177: Am I understanding correctly that now the mean is stashed in the zeroth group? I think it might be useful to comment somewhere explaining this and also maybe add a special enum/int named `MEAN` or something for the group index.

singularity-opac/photons/mean_s_opacity_photons.hpp
```diff
@@ -38,181 +44,467 @@ namespace impl {
...
+  void SetGroupBounds_(const GroupBoundsIndexer &group_bounds,
+                       const int ngroups) {
+    groupBounds_.resize(ngroups + 1);
+    for (int group = 0; group <= ngroups; ++group) {
+      groupBounds_(group) = GroupBoundAt_(group_bounds, group);
+    }
+  }
+
+  template <typename SampleOp>
+  void ForEachGroupFrequencySample_(const Real temp, const Real nuMin,
```
Yurlungur: I wonder if some of this machinery shared between mean and scattering opacities could be unified in some utility functions?

singularity-opac/photons/mean_s_opacity_photons.hpp
```diff
@@ -38,181 +44,467 @@ namespace impl {
...
+      const Real lnu = lNuMin + (inu + 0.5) * dlnu;
+      const Real nu = fromLog_(lnu);
+      sample_op(nu, nu * dlnu);
+    }
+  }
+
+  void ThermalWeightsAtNu_(const PlanckDistribution<PC> &dist, const Real temp,
+                           const Real nu, Real &B, Real &dBdT) const {
+    const Real x = PC::h * nu / (PC::kb * temp);
+    if (x < 80.) {
```
Yurlungur: Why 80 specifically? Should we make this a named constant?

Yurlungur (approved):
Some nitpicks but my comments are non-blocking.

plan_histories/71.md
pdmullen: "Updated" (😬) the plan in [d7894e3](https://github.com/lanl/singularity-opac/pull/71/commits/d7894e31bca2853e5a65df49d87ec405e5905303).

singularity-opac/photons/mean_opacity_photons.hpp
pdmullen: Happy to revert this change.  In `singularity-eos`, however, I do not think we permit ASCII EOS tables.  So maybe there is more symmetry after this changeset? Flexible here...

test/test_mean_opacities.cpp
pdmullen: Removed in [33103fb](https://github.com/lanl/singularity-opac/pull/71/commits/33103fb2e6e9ca53d54dfc756aae910683051bf6).

singularity-opac/photons/mean_opacity_photons.hpp
pdmullen: Noted in [e0269c0](https://github.com/lanl/singularity-opac/pull/71/commits/e0269c059b5ad93c079ddf515130ad0f89ff8f67).

pdmullen:
> Since the gray MeanOpacity class has been extended to multigroup here, does it make sense to support overloads of the opacity accessors that don't take a group index in the event that there is only one group (i.e. the tables are gray)? Or should that be a concern of the downstream codes?

These exist in this changeset already, I believe.  However, I did have some concern that someone might accidentally use them with a multigroup opacity (ngroups > 1) which would result in the behavior of accessing the zeroth group, not a true frequency integrated opacity.  Therefore, I added a `PORTABLE_REQUIRE` in [33103fb](https://github.com/lanl/singularity-opac/pull/71/commits/33103fb2e6e9ca53d54dfc756aae910683051bf6) such that a user can only make these "gray" calls with a single group (i.e., gray) opacity.

singularity-opac/photons/mean_opacity_photons.hpp
pdmullen: Addressed in [Boundary cases for GroupOfNuImpl_](https://github.com/lanl/singularity-opac/pull/71/commits/1a569dbc707b7285c9f2f785b4e44bbfd4f8a682).

singularity-opac/neutrinos/mean_opacity_neutrinos.hpp
pdmullen: To be honest, it has been so long that I can't remember what motivated this change; I apologize.  I do vaguely recall seeing some weird boundary behavior for some multigroup opacities I constructed.  I suspect midpoint rule is more robust because it never touches the boundary values.  We permit group boundaries of zero and +infinity, and there are guards against their evaluation even with trapezoidal rule, but I wonder if midpoint rule was able to bypass those safeguards producing better data... 

I apologize that I cannot remember here.

Yurlungur: All good. If you remember this working better that's fine by me.

singularity-opac/photons/mean_s_opacity_photons.hpp
pdmullen: Named constant in [40b2728](https://github.com/lanl/singularity-opac/pull/71/commits/40b272838391b73daca5a7ff711098b9d6a2b8b3).

singularity-opac/photons/mean_opacity_photons.hpp
pdmullen: The function call that does not take a group index is intended to only ever be used for gray opacities.  This is the reason for an introduced `PORTABLE_REQUIRE` on `ngroups == 1` in response to an earlier review comment from @RyanWollaeger.  I added additional comments to explain its usage.  For future consideration, do we want to have `singularity-opac` supply an over all frequency integrated opacitiy and/or emissivity even when not working with `ngroups=1`? That could be a subsequent PR.

singularity-opac/photons/mean_opacity_photons.hpp
pdmullen: > however, I do not think we permit ASCII EOS tables. So maybe there is more symmetry after this changeset?

@adamdempsey90 says this may not strictly be true.  Let's add more support for other types of EOS tables in the future.  We may even consider adding OpenPMD as acceptable input file formats.  If we need help converting existing opacity ASCII tables into the new spiner format... let me know 😄

singularity-opac/photons/mean_opacity_photons.hpp
Yurlungur: ah ok I misunderstood. I think it can be useful to have a frequency integrated emissivity/opacity for, e.g., Monte Carlo applications. But it can wait for a later mR.

singularity-opac/photons/mean_s_opacity_photons.hpp
pdmullen: I had Claude separate out some of the shared infrastructure in [92d9299](https://github.com/lanl/singularity-opac/pull/71/commits/92d929946cab6a2953772e04874c549c657d3a2e).  As a rule, however, I gave it the guard rail to only move shared utilities that were near identical.  This means that there is likely more generalization to be had among the codebase, but I leave that to a future PR.

Yurlungur:
:+1: merge when ready

singularity-opac/photons/mean_s_opacity_photons.hpp
```diff
@@ -12,207 +12,366 @@
...
-    return std::log10(std::abs(x) + EPS);
-  }
-  PORTABLE_INLINE_FUNCTION Real fromLog_(const Real lx) const {
-    return std::pow(10., lx);
+
+  PORTABLE_INLINE_FUNCTION
+  int GroupOfNuImpl_(const Real nu) const {
+    if (nu == GroupBoundAt(groupBounds_, ngroups_)) {
+      return ngroups_ - 1;
+    }
```
RyanWollaeger: It looks like this is now inconsistent with the absorption check for the low group bound (sorry). No strong opinion (maybe group indexing functions should be unified between absorption and scattering at some point).

singularity-opac/photons/non_cgs_photons.hpp
```diff
@@ -242,79 +243,91 @@ class MeanNonCGSUnits {
...
-    const Real alpha = mean_opac_.RosselandMeanAbsorptionCoefficient(
-        rho_unit_ * rho, temp_unit_ * temp);
-    // alpha output in units of 1/cm. Want to convert out of CGS.
-    // multiplication by length_unit converts length to cm.
-    // division converts length from cm to unit system.
-    // thus multiplication converts (1/cm) to unit system.
-    return alpha * length_unit_;
+  Real RosselandGroupAbsorptionCoefficient(const Real rho, const Real temp,
+                                           const int group) const {
+    return AbsorptionCoefficient(rho, temp, group, Rosseland);
```
RyanWollaeger, line 289: This is a nice trick

RyanWollaeger (approved):
LGTM - one minor residual question, but feel free to ignore.

singularity-opac/photons/mean_s_opacity_photons.hpp
pdmullen: Good catch! Addressed in [8a250d6](https://github.com/lanl/singularity-opac/pull/71/commits/8a250d65ffde057a4e25faa4058cfeb63bb97c4f).

PR #72, opened 2026-06-01
added comments to one test

xevra:
Added some comments.

### DIFF
--- a/test/test_mean_opacities.cpp
+++ b/test/test_mean_opacities.cpp
@@ -446,27 +446,42 @@ TEST_CASE("Mean photon opacities", "[MeanPhotons]") {
 
     constexpr Real kappa = 1.e-20;
 
+    // Load analytic Opacity Eos
     photons::Gray opac_host(kappa);
+    // Put it on the GPU
     photons::Opacity opac = opac_host.GetOnDevice();
 
+    // Load a table Opacity
     photons::MeanOpacity mean_opac_host = photons::MeanOpacityBase(
         opac_host, lRhoMin, lRhoMax, NRho, lTMin, lTMax, NT);
+    // Put it on the GPU
     auto mean_opac = mean_opac_host.GetOnDevice();
 
+    // Declare this thing that lives in a scope where the GPU code is being
+    // managed.
     THEN("The emissivity per nu omega is consistent with the emissity per nu") {
       int n_wrong_h = 0;
 #ifdef PORTABILITY_STRATEGY_KOKKOS
+      // Create a view to avoid copying the data
       Kokkos::View<int, atomic_view> n_wrong_d("wrong");
 #else
+      // Create something that works like a View but is fine on CPU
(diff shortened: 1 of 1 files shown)


test/test_mean_opacities.cpp
```diff
@@ -446,27 +446,42 @@ TEST_CASE("Mean photon opacities", "[MeanPhotons]") {
...
     photons::Opacity opac = opac_host.GetOnDevice();
 
+    // Load a table Opacity
     photons::MeanOpacity mean_opac_host = photons::MeanOpacityBase(
         opac_host, lRhoMin, lRhoMax, NRho, lTMin, lTMax, NT);
+    // Put it on the GPU
     auto mean_opac = mean_opac_host.GetOnDevice();
 
+    // Declare this thing that lives in a scope where the GPU code is being
+    // managed.
```
Yurlungur: this comment is not correct. The THEN macro is just flow control for catch2. It's to make the output of the test executable more readable. See:
https://github.com/catchorg/Catch2/blob/devel/docs/test-cases-and-sections.md#bdd-style-test-cases

test/test_mean_opacities.cpp
```diff
@@ -446,27 +446,42 @@ TEST_CASE("Mean photon opacities", "[MeanPhotons]") {
...
     THEN("The emissivity per nu omega is consistent with the emissity per nu") {
       int n_wrong_h = 0;
 #ifdef PORTABILITY_STRATEGY_KOKKOS
+      // Create a view to avoid copying the data
```
Yurlungur: this will copy data and works on GPU. Kokkos views are not shallow. They're Kokkos's device-capable multidimensional array structure. This creates a device-side view with the atomic trait, which means operations on it are all atomic. This lets us easily (but slowly!) reduce over the number of test failures.

test/test_mean_opacities.cpp
```diff
@@ -446,27 +446,42 @@ TEST_CASE("Mean photon opacities", "[MeanPhotons]") {
...
       Kokkos::View<int, atomic_view> n_wrong_d("wrong");
 #else
+      // Create something that works like a View but is fine on CPU
       PortableMDArray<int> n_wrong_d(&n_wrong_h, 1);
```
Yurlungur, line 469: yeah. This is just a stupid trick to make the syntax identical when Kokkos is disabled.

test/test_mean_opacities.cpp
```diff
@@ -446,27 +446,42 @@ TEST_CASE("Mean photon opacities", "[MeanPhotons]") {
...
 #endif
 
+      // This is a for loop that can be device side or host side
```
Yurlungur: yes, but default is device side.

Yurlungur (changes requested):
Thanks for adding the additional documentation. These aren't all entirely accurate. See my comments below.

Yurlungur (approved):

PR #74, opened 2026-06-02
Cleanup build and tests

Yurlungur:
In this MR, I move from catch2 v2 to catch2 v3 and add a missing include guard in the cmake for spiner. This makes the code/tests more inline with singularity-eos and also makes the cmake actually function for codes that use singularity-opac as a submodule (at least on CPU). The build system still needs updating to be more inline with how singularity-eos does things, but this is a start.

Generative AI was used to assist in the migration.


RyanWollaeger (approved):
LGTM

